PR #275, opened 2018-03-02
Ramps in make_pulses

jpdehollain:
@eendebakpt @lucblom I'm not sure how far along you are with the pulse library, so I'll leave this up to you if you think it's worth merging or not.
The filtering wasn't giving me enough control over the adiabaticity of the pulses in `make_pulses` so I added the option to make ramps of configurable length and relative voltage distance between pulses.
The `proportions` key is not very intuitive or easy to explain in a simple way in the docs, but it was the easiest implementation I could come up with. Open to suggestions if you guys think there's a more user friendly way of doing it.


lucblom:
If you want we can add this feature to the new virtual AWG. @peendebak @jpdehollain Can we close this PR or do we still want to add it to the old virtual AWG?

jpdehollain:
I've been using this with the new awg, and I don't think anyone else is using it, so I'm fine with closing it.

PR #421, opened 2018-08-31
PAT analysis notebook

jpdehollain:
I added some comments to make things clearer and added labels, etc. to the plots.

There was also an odd line in `pat_fitting.pre_process_pat` that I commented out. @CJvanDiepen can you please check if it's ok to remove it? Also, can you please have a look at the `detect_peaks` function and the last section of the example where it is used? there's still a question whether the period is needed, and currently the function is returning an empty dictionary which I don't understand. It might also be good to somehow show the weights in the example.

@peendebak


CJvanDiepen:
@jpdehollain Taking out the commented out line which overwrites `y_data` in `pre_process_pat` looks fine to me, especially if the example notebook looks good after it. 

I cleaned up the last cell in the jupyter notebook where the `detect_peaks` functions is being demonstrated.

I adjusted the `weights` in `detect_peaks` such that they are correct. However they are not being used in the `fit_pat` function, hence I would not add them to the documentation in the jupyter notebook. Removing them completely would be a pity, because in principle assigning weights to the detected peaks looks like a good approach to me, hence they might be used again, possibly after adjusting the way there are calculated, in the future.

CJvanDiepen:
@peendebak @jpdehollain That sounds good to me as well!

PR #422, opened 2018-08-31
Autofit to anticrossing example

jpdehollain:
Mainly worked on the 'Detailed steps' section that was giving a bad fit.

@peendebak can you please check that the lines that are commented out are indeed not needed and can be removed?
Also, I think that there should be an explanation of the outputs of `fit_anticrossing`. Perhaps in a text cell under `print(fit_results)`, and also in the documentation of the function.


peendebak:
@jpdehollain The output of fit_anticrossing should indeed be documented (at least the main fields of the output dictionary). I renamed the output arguments, that should help make the output more readable. I added some docs.
The detailed steps section is maybe not that important. A regular user should not need that part, so we might even remove it from the docs.

PR #425, opened 2018-09-03
RTS example review

jpdehollain:
I swapped the random dataset generation with a real dataset I took. And I did some minor changes to the `random_telegraph_signal`, for aesthetics and to avoid some warnings I was getting with the dataset I loaded.

@peendebak The only issue I still have is that I added a dataset to the `exampledata` folder locally, but git is not recognising it, so I'm not sure how to push it.

@fvanriggelen


peendebak:
@jpdehollain @fvanriggelen Added function to generate RTS signal (used in testing now). Did an autopep

peendebak:
@fvanriggelen @jpdehollain Can you add a short description at the top of the notebook? And perhaps also reference a paper/wiki where RTS is explained.

PR #428, opened 2018-09-04
Examples/fit ohmic2

azwerver:
@peendebak A couple of small changes made to the ohmic fit notebook and a couple of 'spelling mistakes' (e.g. removed ydata=ydata) changed in the script


peendebak:
@azwerver The PR testing fails, see https://travis-ci.com/QuTech-Delft/qtt/builds/83722549. Can you fix the test?

peendebak:
@azwerver The testing fails because the fitting was a little bit unstable. I reduced the noise in the test, which helps and added an option to use lmfit (that seems to work better). Use tested, I want to set the lmfit option to the default.

PR #430, opened 2018-09-04
Pinchoff data was not in folder, so added new data. Changed the folde…

azwerver:
…r path in the notebook

@peendebak Looks nice! The only thing was that the folder where the pinchoff data was supposed to be was removed. I also couldn't find the specific dataset on the K-drive anymore (2018-07-17/12-49-22_qtt_scan1D). So I made an analysis of 100 pinchoff scans and picked the one that I liked best. You can see if you like this scan, if not, feel free to take a different scan.

Also, the pinchoff value is now determined by the midpoint minus 50 mV. This might be totally different for different types of devices, right? It seems a bit arbitrary to me?

Question: in the result notebook, there is a key 'midvalue'. This seems to be the value of the current at the pinchoff point. The name suggest to me, however, that it is the value of the current at the 'midpoint'?


peendebak:
@azwerver Nice catch. Not sure about the midvalue, but for sure the pinchvalue was misleading. I corrected this (see my  commit latest). The `pinchvalue` field I will remove. 

The offset by 50 mV is definitely something we do not want. There is no easy solution though. So I think I will let it be for now and wait unless your work for intel results in a better measure for the pinchoff point.

azwerver:
@peendebak Nice, I just tested it and it looks nicer this way.
I agree with you on the '-50' for the pinchoff point. It's an ongoing process and I'll keep you posted.

PR #441, opened 2018-09-07
examples/polFit

azwerver:
@peendebak @CJvanDiepen 
I changed a bit of the documentation. Furthermore it looks very nice

Two things:

- some constants are defined (h, k), which is fine. It might be more exact to use scipy.constants. I could change that, but only if you want to
- I haven't extensively checked the whole code. Since JP already checked PAT and this code is used many times in the lab. If you think that is needed, I will of course.


eendebakpt:
In general i am in favor of scipy.constants

Op vr 7 sep. 2018 15:48 schreef AM <notifications@github.com>:

azwerver:
I'll have a look at it later today

PR #442, opened 2018-09-07
Fit Ramsey review

jpdehollain:
Fixed title, descriptions and modified the parameters in the fit function to make it clear that the decay is T2*

@peendebak


peendebak:
@jpdehollain  The new notebook looks good. I like the rename of `C` into `1/t2s`. The change is not backwards compatible though. So @xuexiao1992 @fvanriggelen check whether you use this function, and if you do, adapt your code.

PR #452, opened 2018-09-11
2x2 tuning example

jpdehollain:
Made an example notebook with how I use `dotsystem` to run charge stability diagram simulations of the 2x2.

@peendebak


lucblom:
Some review comments:

- Do not abbreviate variable names (so e.g. `dot_system` and not `ds`). What is the full name of `hcgs`, `det` and `la_matrix`?

- Changing function argument variables in a function is not really a good idea. Only use the variables. For example don't do something like:
```
def my_function(dot_system):
   dot_system.some_variable = 2
```

- How are the `det_to_plunger` values determined?

- I'm not sure what the `parse_scan_parameters` is doing. Can you explain?

- I updated the `clickOCC` function. Note it is much more readable for other users now.

Can we quickly discuss and finish up the PR? @peendebak mentioned a beter location for the example notebook.

peendebak:
@lucblom @jpdehollain I updated to example and added it to #472. I did check the resulting charge stability diagram looks the same, but did not do more testing

Issue #439, opened 2018-09-06
use qupulse instead of qctoolkit


peendebak:
The qctoolkit package was renamed to qupulse. The code need to be updated to reflect this (mainly the new virtual awg and the setup.py/requirements.txt/dev-requirements.txt).

@lucblom

PR #455, opened 2018-09-12
Moved qctoolkit references to qupulse.

lucblom:
@peendebak 

Should resolve the issue #439. Currently the master branch of qupulse gives a warning when not using `fractions.Fraction` as [you](https://github.com/qutech/qupulse/issues/375) mentioned.


peendebak:
Looks good. I think qtt is already off jenkins. So 💃

PR #456, opened 2018-09-16
improve documentation for virtual gates object

peendebak:
* Improve documentation
* Use name VirtualDAC instead of virtual_IVVI

@lucblom


peendebak:
@lucblom I suspect the return arguments of `qupulse.pulses.plotting.render` have changed. I had to make a small change to the `sequences.py`.

PR #460, opened 2018-09-19
dev to master

lucblom:
@peendebak 

Maybe we want to define what some rules to merge to master (e.g. update version number)?


peendebak:
Hi Luc,

Let’s agree on the following:


  *   Whenever we merge from dev to master we bump the version in qtt/version.py
  *   When the change is large enough we also create a new pip package. (Ask me how you can do this at some moment)
  *   We try to follow semantic versioning https://semver.org/

And notify if we to something else for one reason or the other.

Pieter

PR #475, opened 2018-10-24
[DEM-524] Improve HDAWG8 virtual awg backend

qSaevar:


qtt/instrument_drivers/virtualAwg/awgs/ZurichInstrumentsHDAWG8.py
```diff
@@ -66,94 +74,37 @@ def retrieve_running_mode(self):
...
+            if sampling_rate == value:
+                self.__awg.set('awgs_{}_time'.format(self.__awg_number), key)
+                return
+        raise ValueError('Sampling rate {} not in available a list of available values: {}'.format(
+            sampling_rate, self.__sampling_rate_map))
 
     def retrieve_sampling_rate(self):
-        #ziDAQ('setInt', '/dev8044/awgs/0/time', 0);
-        return 2.4e9
+        sample_rate = self.__awg.get('awgs_{}_time'.format(self.__awg_number))
```
eendebakpt: If consistent with the name of the function, it should be `sampling_rate` instead of `sample_rate`. Also, it is not the sampling rate, but rather an index. So maybe `sampling_rate_index` or `sampling_rate_key`

qtt/instrument_drivers/virtualAwg/awgs/ZurichInstrumentsHDAWG8.py
```diff
@@ -66,94 +74,37 @@ def retrieve_running_mode(self):
...
     def update_gain(self, gain):
-        raise NotImplementedError
+        [self.__awg.set('sigouts_{}_range'.format(ch), gain) for ch in self._channel_numbers]
 
     def retrieve_gain(self):
-        # ziDAQ('setDouble', '/dev8044/sigouts/0/range', 1);
-        return 1
+        gains = [self.__awg.get('sigouts_{}_range'.format(ch)) for ch in self._channel_numbers]
+        if not all(g == gains[0] for g in gains):
+            raise ValueError('Not all channel amplitudes are equal. Please reset!')
```
eendebakpt: The function name has `retrieve`, the call to `__awg` has `range` and the exception raised `amplitude`. There can be only one! The retrieve is mandated by the `AwgCommon`, but at least rename the amplitudes to either range or gain.
Or do you really mean amplitude?

qtt/tests/test_zi_hdawg8.py
```diff
@@ -0,0 +1,88 @@
...
+    def test_enable_outputs(self):
+        self.zi_hdawg8.enable_outputs()
+        calls = [call.enable_channel(ch) for ch in range(0, 8)]
+        self.awg.assert_has_calls(calls)
+
+        with self.assertRaises(AwgCommonError):
+            self.zi_hdawg8.enable_outputs([0, 1, 2, 3, 8])
+
+        self.zi_hdawg8.enable_outputs([6, 7])
+        calls + [call.enable_channel(ch) for ch in range(6, 7)]
```
eendebakpt:
```suggestion
        calls += [call.enable_channel(ch) for ch in range(6, 7)]
```

qtt/tests/test_zi_hdawg8.py
```diff
@@ -0,0 +1,88 @@
+import unittest
+from unittest.mock import MagicMock, call
+
+from qtt.instrument_drivers.virtualAwg.awgs.ZurichInstrumentsHDAWG8 import ZurichInstrumentsHDAWG8
+from qtt.instrument_drivers.virtualAwg.awgs.common import AwgCommonError
+
+
+class TestZurichInstrumentsHDAWG8(unittest.TestCase):
```
eendebakpt: Missing test cases for `run`, `stop`, `delete_waveforms`, ... It does mean that coverage of the ZurichInstrumentsHDAWG8 will not be 100%.
Since this is in `qtt` I am fine with not doing 100%.

qtt/tests/test_zi_hdawg8.py
```diff
@@ -0,0 +1,88 @@
...
+    def test_disable_outputs(self):
+        self.zi_hdawg8.disable_outputs()
+        calls = [call.disable_channel(ch) for ch in range(0, 8)]
+        self.awg.assert_has_calls(calls)
+
+        with self.assertRaises(AwgCommonError):
+            self.zi_hdawg8.disable_outputs([0, 1, 2, 3, 8])
+
+        self.zi_hdawg8.disable_outputs([6, 7])
+        calls + [call.disable_channel(ch) for ch in range(6, 7)]
```
eendebakpt:
```suggestion
        calls += [call.disable_channel(ch) for ch in range(6, 7)]
```

qtt/instrument_drivers/virtualAwg/awgs/ZurichInstrumentsHDAWG8.py
```diff
@@ -1,63 +1,71 @@
...
+        super().__init__('ZIHDAWG8', channel_numbers=list(range(0, 8)),
+                         marker_numbers=list(range(0, 8)))
         if type(awg).__name__ is not self._awg_name:
             raise AwgCommonError('The AWG does not correspond with {}'.format(self._awg_name))
-        self.__device_name = awg._dev.device
-        self.__awg_module = awg._dev.awgModule
-        self.__daq = awg._dev.daq
         self.__awg = awg
+        self.__awg_number = awg_number
+        self.__sampling_rate_map = {0: 2.4e9, 1: 1.2e9, 2: 600e6, 3: 300e6, 4: 150e6, 5: 72e6, 6: 37.50e6, 7: 18.75e6,
```
eendebakpt: This could be a static variable

qtt/tests/test_zi_hdawg8.py
```diff
@@ -0,0 +1,88 @@
...
+        for sample_rate in sample_rates:
+            self.zi_hdawg8.update_sampling_rate(sample_rate)
+        calls = [call.set('awgs_0_time', i) for i in range(0, 14)]
+        self.awg.assert_has_calls(calls)
+
+        with self.assertRaises(ValueError):
+            self.zi_hdawg8.update_sampling_rate(99)
+
+    def test_retrieve_sampling_rate(self):
+        self.awg.get.return_value = 5
```
eendebakpt: The relation between 5 and the 72e6 on the line below is not clear from the unit test alone. Better perhaps something like:
```
sampling_rate_index=5
self.awg.get.return_value = sampling_rate_index
expected_sampling_rate=self.sample_rates[sampling_rate_index]
self.assertEqual(expected_sampling_rate, self.zi_hdawg8.retrieve_sampling_rate())
```

qtt/tests/test_zi_hdawg8.py
```diff
@@ -0,0 +1,88 @@
...
+        self.zi_hdawg8.update_gain(0.5)
+        calls = [call.set('sigouts_{}_range'.format(ch), 0.5) for ch in range(8)]
+        self.awg.assert_has_calls(calls)
+
+    def test_retrieve_gain(self):
+        self.awg.get.return_value = 0.2
+        self.assertEqual(0.2, self.zi_hdawg8.retrieve_gain())
+
+        with self.assertRaises(ValueError):
+            self.awg.get.side_effect = lambda v: v
```
eendebakpt: What does this mean? `self.awg` is a magic mock object. But what does `.get.side_effect` do?

qSaevar: with the lambda function it just returns the value that is used, in this case:
`sigouts_{}_range`
So in retrieve_gain:
`gains = [self.__awg.get('sigouts_{}_range'.format(ch)) for ch in self._channel_numbers]`
the gains is just the list of channel numbers and a ValueError will be raised as not all gains are equal.

PR #479, opened 2018-11-20
dataviewer: Menu and filter

dpfranke:
Added menu to clean up GUI, added filename filter

@lucblom


lucblom:
@dpfranke 

Works but would like to clean up a bit. Will merge.

PR #488, opened 2018-12-21
Add detectCoulombPeaks which directly works an arrays of data.

CJvanDiepen:
@eendebakpt

### DIFF
--- a/qtt/algorithms/chargesensor.py
+++ b/qtt/algorithms/chargesensor.py
@@ -121,7 +121,7 @@ def correctChargeSensor(xscan, yscan, xs, ys, fig=None):
 
     """
 
-    peaks = qtt.algorithms.coulomb.coulombPeaks(xs, ys, fig=fig, istep=1)
+    peaks = qtt.algorithms.coulomb.coulombPeaks(xs, ys, fig=fig, sampling_rate=1)
 
     try:
         # range for fit: select from the first detected peak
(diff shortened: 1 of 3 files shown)


peendebak:
@lucblom @CJvanDiepen Adding another function that works directly on numpy arrays is fine I think and the implementation looks good. I am a bit worried that there will be confusion between the `detectCoulombPeaks` and `analyseCoulombPeaks` and think we should not merge the current PR. 

Some options:
* Make a `analyseCoulombPeaks(dataset, ...)` and a `_analyseCoulombPeaks(xdata, ydata, ...)`
* Make a `analyseCoulombPeaks(dataset, ...)` and a `analyseCoulombPeaks???(xdata, ydata, ...)`
* Overload the `analyseCoulombPeaks` to `analyseCoulombPeaks(dataset=None, xdata=None, ydata=None, ...)`
* Add another argument to `analyseCoulombPeaks` that specifies the `DataArray` to select from the `DataSet`. E.g. `analyseCoulombPeaks(dataset, ..., data_array='I')`.
* Add a simple function that makes a clone of a `DataSet` while only keeping a single array (e.g. the 'I' or 'Q' array). This is an alternative solution to the problem this PR tries to solve.

Any other suggestions or preference for one of the above options?

eendebakpt:
@CJvanDiepen Ok, let's do option 2 with the name you suggested. Also note that the tests fail because you typed `detectCoulombpeaks` instead of `detectCoulombPeaks`

CJvanDiepen:
@peendebak Renamed so looks ready to me.

Is there a command which I could run to do the tests myself before opening the pull request?

peendebak:
The unittests you can run with `pytest`. The example notebooks can be executed with `docs\notebooks\run_notebooks.sh` (this is for unix, for windows you might need to adapt it a little bit). @CJvanDiepen

PR #489, opened 2018-12-21
Addition line fitting

CJvanDiepen:
@peendebak 

Add fit_add_line_arr to fit an addition line on data in the form of arrays and adjust the initial guess for the center and the slope in initFermiLinear

### DIFF
--- a/qtt/algorithms/fitting.py
+++ b/qtt/algorithms/fitting.py
@@ -53,10 +53,24 @@ def initFermiLinear(x_data, y_data, fig=None):
         # subtract linear part
         yr = ydata - y
 
-        cc = np.mean(xdata)
+        sigma = xdata.size/250
+        dyr = scipy.ndimage.gaussian_filter(yr, sigma, order=1)
+        
+        # assume step is steeper than overall slope
+        cc_idx = np.argmax(np.abs(dyr))            
+        h = int(xdata.size / 2)
+        
+        # prevent guess to be at the edges
+        if cc_idx < xdata.size/100 or cc_idx > (99/100)*xdata.size:
+            cc = np.mean(xdata)
+            A = -(np.mean(yr[h:]) - np.mean(yr[:h]))
+        else:
+            cc = xdata[cc_idx]
+            A = -1*np.sign(dyr[cc_idx])*(np.mean(yr[h:]) - np.mean(yr[:h]))
+        
         h = int(xdata.size / 2)
         A = -(np.mean(yr[h:]) - np.mean(yr[:h]))
-        T = np.std(xdata) / 10
+        T = np.std(xdata) / 100
         ab[1] = ab[1] - A / 2  # correction
         ylin = linear_function(xdata, ab[0], ab[1])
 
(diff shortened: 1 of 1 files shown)


qtt/algorithms/fitting.py
```diff
@@ -53,10 +53,24 @@ def initFermiLinear(x_data, y_data, fig=None):
         # subtract linear part
         yr = ydata - y
 
-        cc = np.mean(xdata)
+        sigma = xdata.size/250
```
eendebakpt: Create a subfuction `_estimate_fermi_model_center(...)` (or something like that) out of lines 56 to (roughly) 70

qtt/algorithms/fitting.py
```diff
@@ -53,10 +53,24 @@ def initFermiLinear(x_data, y_data, fig=None):
...
+        cc_idx = np.argmax(np.abs(dyr))            
+        h = int(xdata.size / 2)
+        
+        # prevent guess to be at the edges
+        if cc_idx < xdata.size/100 or cc_idx > (99/100)*xdata.size:
+            cc = np.mean(xdata)
+            A = -(np.mean(yr[h:]) - np.mean(yr[:h]))
+        else:
+            cc = xdata[cc_idx]
+            A = -1*np.sign(dyr[cc_idx])*(np.mean(yr[h:]) - np.mean(yr[:h]))
```
eendebakpt: You define `A` here, but it overwritten on line 72

qtt/algorithms/fitting.py
```diff
@@ -145,6 +159,32 @@ def fermi_linear_fitting_function(xdata, a, b, cc, A, T):
 
 # %%
 
+def fit_add_line_arr(xdata, ydata, trimborder=True):
```
eendebakpt: This function duplicates code from `fit_addition_line` and I would like to prevent that. Can you use the `fit_add_line_arr` in `fit_addition_line`)?

Also: since `fit_add_line_arr` is a public function, I want a proper name, e.g. `fit_addition_line_array`.

qtt/algorithms/fitting.py
```diff
@@ -53,10 +53,24 @@ def initFermiLinear(x_data, y_data, fig=None):
...
+        dyr = scipy.ndimage.gaussian_filter(yr, sigma, order=1)
+        
+        # assume step is steeper than overall slope
+        cc_idx = np.argmax(np.abs(dyr))            
+        h = int(xdata.size / 2)
+        
+        # prevent guess to be at the edges
+        if cc_idx < xdata.size/100 or cc_idx > (99/100)*xdata.size:
+            cc = np.mean(xdata)
+            A = -(np.mean(yr[h:]) - np.mean(yr[:h]))
```
eendebakpt: Why not
```suggestion
             A = -1*np.sign(dyr[cc_idx])*(np.mean(yr[h:]) - np.mean(yr[:h]))
```
As for the other case?

eendebakpt:
@CJvanDiepen The tests sometimes failed (sometimes because data is generated randomly). I fixed the tests and renamed `fit_add_line_array` to `fit_addition_line_array`

CJvanDiepen:
@eendebakpt 

Code looks good to me. Minor comment is whether or not the if statement in line 256 should have ```verbose``` i.o. ```fig``` as condition since there are only print statements in the if statement. Will you perform the merge?

CJvanDiepen:
@peendebak @lucblom Could you review and merge this PR?

PR #493, opened 2019-01-08
fix rtd build with qupulse

peendebak:
@lucblom


lucblom:
@peendebak 

Did you forgot to add the readthedocs file?
What is the problem with qupulse?

peendebak:
@lucblom The readthedocs file is alreadt there (`.readthedocs.yml`). I am not sure what the issue is, but the build might fail because `qupulse` was not installed. See https://readthedocs.org/projects/qtt/builds/8362858/

PR #494, opened 2019-01-09
Implemented better version of pulse gates.

CumulonimbusCalvus:


qtt/instrument_drivers/virtualAwg/sequencer.py
```diff
@@ -58,6 +58,32 @@ def make_square_wave(amplitude, period, repetitions=1, name='pulse'):
         return {'name': name, 'wave': SequencePT(*(sequence_data,)*repetitions),
                 'type': DataTypes.QU_PULSE}
 
+    @staticmethod
+    def make_pulse_table(amplitudes, waiting_times, repetitions=1, name='pulse_table'):
+        """ Creates a sequence of pulses from a list of amplitudes and waiting times.
+            Note that the initial voltage level will be given by the last element in amplitudes.
```
peendebak: This is really odd. Can we change such that the first value is the first value in the list of amplitudes?

qtt/instrument_drivers/virtualAwg/sequencer.py
```diff
@@ -58,6 +58,32 @@ def make_square_wave(amplitude, period, repetitions=1, name='pulse'):
...
+
+        Args:
+            amplitudes (list of floats): List with voltage amplitudes of the pulses.
```
peendebak: I think the 'official' annotation is `list[float]` (see https://www.python.org/dev/peps/pep-0526/). `list of float` is also fine for me, but mixed with `list of floats`

qtt/instrument_drivers/virtualAwg/sequencer.py
```diff
@@ -58,6 +58,32 @@ def make_square_wave(amplitude, period, repetitions=1, name='pulse'):
...
+            waiting_times (list of float): List with durations containing the waiting time of each pulse.
+            repetitions (int): The number of oscillations in the sequence.
```
peendebak:
```suggestion
            repetitions (int): The number of repetitions of the base sequence.
```
Oscillations is not the correct term

qtt/instrument_drivers/virtualAwg/templates.py
```diff
@@ -60,3 +60,7 @@ def marker(name):
         """
         return TablePT({name: [(0, 0), ('offset', 1),
                                ('offset+uptime', 0), ('period', 0)]})
+
+    @staticmethod
+    def pulse_table(name, entries):
```
peendebak: Missing docstring. The function is so short, do we need it?

qtt/instrument_drivers/virtualAwg/sequencer.py
CumulonimbusCalvus: ✔️

qtt/instrument_drivers/virtualAwg/sequencer.py
CumulonimbusCalvus: ✔️

qtt/instrument_drivers/virtualAwg/templates.py
CumulonimbusCalvus: ✔️

PR #497, opened 2019-01-24
Trigger delay handling of virtual_awg

CJvanDiepen:
@peendebak @CumulonimbusCalvus 

Fix delay for readout trigger by virtual_awg for scan1Dfast and scan2Dfast by passing through the sampling frequency.


qtt/measurements/scans.py
```diff
@@ -494,6 +494,21 @@ def scan1Dfast(station, scanjob, location=None, liveplotwindow=None, delete=True
     t0 = time.time()
     wait_time_startscan = scanjob.get('wait_time_startscan', 0)
 
+    try:
```
CumulonimbusCalvus: You could turn this addition into a (private) function and use it in the `scan1Dfast` and `scan2DFast` function (dry principle).

qtt/measurements/scans.py
```diff
@@ -494,6 +494,21 @@ def scan1Dfast(station, scanjob, location=None, liveplotwindow=None, delete=True
...
+        isfpga = isinstance(
+            minstrhandle, qtt.instrument_drivers.FPGA_ave.FPGA_ave)
+    except:
+        isfpga = False
+    try:
```
CumulonimbusCalvus: Why is the `isinstance` in a `try` and `except` block? If the `minstrhandle` is not of the correct type it will return False. Am I maybe missing something?

CJvanDiepen:
@CumulonimbusCalvus  @eendebakpt 

I agree with both comments by Luc. I think this code was originally written by Pieter and I only copy-pasted it to make the different scan functions consistent with each other. Pieter, could you reply on Luc's comment and implement them as suggested if you agree?

qtt/measurements/scans.py
peendebak: `qtt.instrument_drivers.FPGA_ave.FPGA_ave` cannot be imported on a system without an FPGA, therefore we used a `try`-`catch`. Better would have been to catch the specific exception. @CumulonimbusCalvus

peendebak:
@CJvanDiepen How is the delay fixed with this PR? I see the sampling frequency is added to many functions of the virtual awg, but I do not see it being used.

Most of the changes are related to the old virtual awg object. We will probably stop supporting that in the near future, so I suggest not to put a large amount of work into this. (It does also mean I will not block a PR that is not perfect)

Making a subfunction to be reused by the different scan functions should not be that hard. Something like:
```
from qtt.measurements.scans import *

def sample_rate_from_instrument_handle(measurement_instrument_handle):
    """ Retrieve current sampling rate from an instrument
    
    Args:
        measurement_instrument_handle (tuple or str or Instrument): specification of the instrument
    Returns:
        float: current sampling rate
    """
    
    instrument = get_instrument(measurement_instrument_handle)
    
    try:
        isfpga = isinstance(
            instrument, qtt.instrument_drivers.FPGA_ave.FPGA_ave)
    except:
        isfpga = False
    try:
        ism4i = isinstance(
            instrument, qcodes.instrument_drivers.Spectrum.M4i.M4i)
    except:
        ism4i = False
    if isfpga:
        sample_rate = instrument.get_sampling_frequency()
    elif ism4i:
        sample_rate = instrument.sample_rate()
    return sample_rate
```
(I did not test this, just to give an idea)

CJvanDiepen:
@peendebak The pretrigger_period is set based on the samp_freq. Hence, this introduces a conditional delay in the data acquisition. For some sweep functions in the virtual_awg this samp_freq argument was introduced but for others it apparently was not. This PR adds it to the ones still missing it, such that the pretrigger_period is set consistently. 

Pieter, your suggestion for a subfunction to obtain the sampling frequency is then perhaps a good idea for the new virtual_awg instrument? (@CumulonimbusCalvus )

CumulonimbusCalvus:
@CJvanDiepen Can you test the scan updates I made on your setup, to see whether I did not made any mistake?

peendebak:
@CJvanDiepen @CumulonimbusCalvus This PR has many conflicts with current dev. Is the functionality still required? If so, then I suggest we open a new PR.

PR #498, opened 2019-01-24
Allow for stepvalues usage for MultiParameter handling in scan2Dfast

CJvanDiepen:
@eendebakpt @CumulonimbusCalvus 

Add usage of stepvalues for operation of MultiParameter object. Used for example to step both a mw power and frequency via the stepdata of the scan2Dfast.

### DIFF
--- a/qtt/measurements/scans.py
+++ b/qtt/measurements/scans.py
@@ -1843,7 +1843,15 @@ def scan2Dfast(station, scanjob, location=None, liveplotwindow=None, plotparam='
     t0 = time.time()
 
     if type(stepvalues) is np.ndarray:
-        stepvalues_tmp = stepdata['param'][list(stepvalues[:, 0])]
+        if stepvalues.ndim > 1:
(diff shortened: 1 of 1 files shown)


qtt/measurements/scans.py
```diff
@@ -1904,6 +1912,16 @@ def scan2Dfast(station, scanjob, location=None, liveplotwindow=None, plotparam='
         liveplotwindow.update_plot()
         pyqtgraph.mkQApp().processEvents()
 
+    # when using multiparameter object as stepdata param add stepped values to dataset for all step parameters
+    if hasattr(stepvalues, 'ndim'):
```
CumulonimbusCalvus:
```suggestion
    if hasattr(stepvalues, 'ndim') and stepvalues.ndim > 1:
```

CumulonimbusCalvus: Only a small suggestion to reduce indentation.

CumulonimbusCalvus: @CJvanDiepen I fixed it for you.

PR #499, opened 2019-01-27
fix bug in initial estimate of fermi fitting

eendebakpt:
@CumulonimbusCalvus @CJvanDiepen 

Fix issue with `_estimate_fermi_model_center` (which I renamed to `_estimate_fermi_model_center_amplitude`)

### DIFF
--- a/qtt/algorithms/fitting.py
+++ b/qtt/algorithms/fitting.py
@@ -1,5 +1,7 @@
 """ Fitting of Fermi-Dirac distributions. """
 
+import warnings
+
 import matplotlib.pyplot as plt
 import numpy as np
 import scipy
@@ -9,7 +11,7 @@ from qcodes import DataArray
 from qtt.algorithms.functions import Fermi, FermiLinear, linear_function
 
 
-def _estimate_fermi_model_center(x_data, y_data_linerized):
+def _estimate_fermi_model_center_amplitude(x_data, y_data_linearized, fig=None):
     """ Estimates the following properties of a charge addition line; the center location
         of the addition line. The amplitude step size caused by the addition line.
 
@@ -22,10 +24,10 @@ def _estimate_fermi_model_center(x_data, y_data_linerized):
         amplitude_step (float): Estimate of the amplitude of the step.
     """
     sigma = x_data.size / 250
-    y_filtered = scipy.ndimage.gaussian_filter(y_data_linerized, sigma, order=1)
+    y_derivative_filtered = scipy.ndimage.gaussian_filter(y_data_linearized, sigma, order=1)
 
     # assume step is steeper than overall slope
-    estimated_index = np.argmax(np.abs(y_filtered))
+    estimated_index = np.argmax(np.abs(y_derivative_filtered))
     center_index = int(x_data.size / 2)
 
     # prevent guess to be at the edges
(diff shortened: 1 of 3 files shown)


CJvanDiepen:
@eendebakpt 

The variable namings are indeed improved like this. Seems that I somehow messed up using both the normal data and the derivative data in the previous adjustment. Thanks for fixing this. 

In the function ```_estimate_fermi_model_center_amplitude``` the usage of both ```step_size``` and ```amplitude_step``` seems redundant, hence one can be removed.

peendebak:
@CJvanDiepen @CumulonimbusCalvus Done

qtt/algorithms/functions.py
```diff
@@ -233,7 +233,7 @@ def exp_function(x, a, b, c):
     return y
 
 
-def cost_exp_decay(x_data, y_data, params, threshold=None):
+def cost_exp_decay(x_data, y_data, params, threshold=None, offset_parameter = None):
```
CumulonimbusCalvus: There is an argument `params` which can set the `a` argument of `exp_function`. Therefore I do not think the `offset_parameter` argument is needed.

I think you can do use the `params` in a more fancy way, by making use of `**kwargs`.

qtt/algorithms/fitting.py
```diff
@@ -34,11 +36,33 @@ def _estimate_fermi_model_center(x_data, y_data_linerized):
...
-    amplitude_step = -1.0 * np.sign(y_filtered[estimated_index]) * step_size
+    split_offset = int(np.floor(x_data.size/10))
+    mean_right = np.mean(y_data_linearized[(center_index+split_offset):])
+    mean_left = np.mean(y_data_linearized[:(center_index-split_offset)])
+    amplitude_step = -(mean_right - mean_left)
+    
+    if np.sign(-y_derivative_filtered[estimated_index]) != np.sign(amplitude_step):
+        warnings.warn('step size might be incorrect')
+        
+    if fig is not None:
```
CumulonimbusCalvus: I think it would be more readable if you move the plotting to a separate function.

qtt/algorithms/functions.py
peendebak: Fair point. I came up with a solution. To test the speed I used the following:
```
import numpy.core.numeric
import numpy as np
from qtt.algorithms.functions import *

params=[2., 3.]
offset_parameter=2.2

x_data=np.random.rand()
y_data=exp_function(x_data, offset_parameter, *params)+np.random.rand()

threshold=1.

#%%

def func1(params): return cost_exp_decay(x_data, y_data, params, threshold, None)

print(func1([offset_parameter,1,1]))

# %timeit func1([offset_parameter, 1,1]): 8,5 us

#%% Current method
def func2(params): return cost_exp_decay(x_data, y_data, params, threshold, offset_parameter)

print(func2([1,1]) )

# %timeit func2([1,1]): 10.2 us 

#%% Alternative method (would simplify cost_exp_decay since offset_parameter is not needed any more)



def func3(params): return cost_exp_decay(x_data, y_data, np.hstack( (offset_parameter, params)), threshold)

# not much faster
#def func3(params): return cost_exp_decay(x_data, y_data, numpy.core.numeric.concatenate(([offset_parameter], np.array(params) ), 0), threshold)
# fast but not robust for all input types
#def func3(params): return cost_exp_decay(x_data, y_data, [offset_parameter]+params, threshold)

print(func3([1,1]) )
```
The third options does what we want, but it quite a bit. Do you see alternatives?

@CumulonimbusCalvus

peendebak:
@CumulonimbusCalvus Addressed the comments

PR #502, opened 2019-01-31
Fix AWG to plunger

CumulonimbusCalvus:
@fvanriggelen 

Me and @CJvanDiepen converted the AWG to plunger conversion factor back in the new virtual AWG, since it was making matters not more clear. Just a heads up, since you have been using the virtual AWG as well.

### DIFF
--- a/docs/notebooks/measurements/example_virtual_awg.ipynb
+++ b/docs/notebooks/measurements/example_virtual_awg.ipynb
@@ -43,7 +43,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


peendebak:
@CumulonimbusCalvus @CJvanDiepen This basically changes the AWG_to_plunger to 1/AWG_to_plunger with an added factor of 1e3. Even though 1/AWG_to_plunger makes more sense, I suggest to choose a new name altogether to avoid confusion and conversion issues (e.g. `awg_to_plunger`, `awg_channel_to_gate`, `awg_to_gate` or similar)

CumulonimbusCalvus:
@peendebak 

This PR changes the `awg_to_plunger` into exactly the same conversion factor as in the old virtual AWG. Does it then still make sense to rename the parameter?

CJvanDiepen:
@CumulonimbusCalvus @peendebak By re-defining it in the new virtual_awg as proposed in this PR the ```awg_to_plunger``` indeed is the same as it was in the old virtual_awg. Hence renaming it is not necessary.

Perhaps renaming it to ```awg_to_gate```, because the gate does not have to be a plunger would still be nice.

PR #504, opened 2019-02-05
Misc updates from qi fridge

peendebak:


qtt/instrument_drivers/virtualAwg/virtual_awg.py
```diff
@@ -420,7 +420,9 @@ def sequence_gates(self, sequences, do_upload=True):
             upload_data.append((sequence_names, sequence_channels, sequence_items))
             if do_upload and sequence_items:
                 self.awgs[number].upload_waveforms(sequence_names, sequence_channels, sequence_items)
-        return {'gate_comb': sequences, 'upload_data': upload_data, 'settings': settings_data}
+                
+        self._sequence_data = {'gate_comb': sequences, 'upload_data': upload_data, 'settings': settings_data}
```
CumulonimbusCalvus: OK???

peendebak: After performing an action with the virtual awg (e.g. `sweep_gates`) I want to be able te retrieve the data that was uploaded (for debugging). There data is returned by the `sequence_gates`, but the calling method does not always store that data. Would it help if I create a public method `virtual_awg.return_latest_sequence_data()`?

@CumulonimbusCalvus

peendebak:
> I think plotting should not be part of the scan functionality (e.g. scan1DFast). The plotting should be separated or implemented with a generic callback.
> 
> You now need the additional argument `plotparam` (and in future more) to change the settings of the plotting. This is not very maintainable and future prove.
> 
> Can we change this now or at a later stage?

@CumulonimbusCalvus This is not for the full plotting functionality, but for the live plotting. An alternative would be to create a global `LivePlotting` object and let the `scan1Dfast` retrieve the required information. A session would then be something like:
```
# no arguments, live plotting will use `.default_parameter_array` for plotting
dataset = scan1Dfast(...)

LivePlotting.set_plot_parameters(['amplitude_1', 'amplitude_2']
# now the live plotting will plot the arrays set on the line before this
dataset = scan1Dfast(...)
```
The live plotting ties closely to the measurement loops, so I am not sure how much time to invest here.

peendebak:
@CumulonimbusCalvus Check commit f3635a8 please. I am not sure why this was not picked up before, but basically a
``` 
scanjob['sweepgates']['param']='SDP'
``` 
and
``` 
scanjob['sweepgates']['param']={'SDP': 1, 'P1': 0}
``` 
gave different results, while they should give the same results.

qtt/instrument_drivers/virtualAwg/awgs/ZurichInstrumentsHDAWG8.py
```diff
@@ -4,7 +4,7 @@
 
 class ZurichInstrumentsHDAWG8(AwgCommon):
     __sampling_rate_map = {0: 2.4e9, 1: 1.2e9, 2: 600e6, 3: 300e6, 4: 150e6, 5: 72e6, 6: 37.50e6, 7: 18.75e6,
-                           8: 9.4e6, 9: 4.5e6, 10: 2.34e6, 11: 1.2e3, 12: 586e3, 13: 293e3}
+                           8: 9.4e6, 9: 4.5e6, 10: 2.34e6, 11: 1.2e6, 12: 586e3, 13: 293e3}
```
eendebakpt: @CumulonimbusCalvus Was this indeed a typo?

peendebak:
@CumulonimbusCalvus Tests are green again. It was indeed related to gui elements running on travis.

qtt/instrument_drivers/virtualAwg/awgs/ZurichInstrumentsHDAWG8.py
CumulonimbusCalvus: Yes, but there where some more sample_rate value typo's. I updated the `sampling_rate_map` to the correct values.

PR #509, opened 2019-02-26
[DEM-793] Allow negative trigger delays

qSaevar:


qtt/instrument_drivers/virtualAwg/sequencer.py
```diff
@@ -98,17 +99,33 @@ def make_marker(period, uptime=0.2, offset=0.0, repetitions=1, name='marker'):
...
-            raise ValueError('Invalid argument value (uptime <= 0 or offset < 0)!')
-        if uptime + offset > period:
-            raise ValueError('Invalid argument value (uptime + offset > period)!')
-        input_variables = {'period': period * Sequencer.__sec_to_ns,
-                           'uptime': uptime * Sequencer.__sec_to_ns,
-                           'offset': offset * Sequencer.__sec_to_ns}
-        sequence_data = (Templates.marker(name), input_variables)
-        return {'name': name, 'wave': SequencePT(*((sequence_data,)*repetitions)),
+
+        if abs(offset) > period:
```
eendebakpt: Perhaps move these two checks into the `__calculate_pulse_parameters`?

qtt/instrument_drivers/virtualAwg/sequencer.py
```diff
@@ -98,17 +99,33 @@ def make_marker(period, uptime=0.2, offset=0.0, repetitions=1, name='marker'):
...
+            raise ValueError('Invalid argument value for offset: |{}| > {}!'.format(offset, period))
+        if not 0 < uptime < period:
+            raise ValueError("Invalid argument value for uptime '{}'!".format(uptime))
+        input_variables = Sequencer.__calculate_pulse_parameters(offset, period, uptime)
+        sequence_data = (Templates.rollover_marker(name), input_variables)
+        return {'name': name, 'wave': SequencePT(*((sequence_data,) * repetitions)),
                 'type': DataTypes.QU_PULSE, 'uptime': uptime, 'offset': offset}
 
+    @staticmethod
+    def __calculate_pulse_parameters(offset, period, uptime):
```
eendebakpt: The name is not descriptive enough. The method is really about the marker parameters and not a general pulse. Maybe `_calculate_marker_pulse_parameters`?

qtt/instrument_drivers/virtualAwg/templates.py
```diff
@@ -60,3 +60,19 @@ def marker(name):
         """
         return TablePT({name: [(0, 0), ('offset', 1),
                                ('offset+uptime', 0), ('period', 0)]})
+
+    @staticmethod
+    def rollover_marker(name):
+        """Creates a TTL pulse qupulse template for sequencing.
```
eendebakpt: I understand the parameters of this pulse template, because I just read the (private) `__calculate_pulse_parameters`. Without that I would have a hard time understanding what kind of pulse this is. The usage of the boolean `rollover` variable is particularly "interesting".

The parameters for this pulse are generated from `__calculate_pulse_parameters` which is ok, but if this was a pulse template to be used by a user, I would like a different interface (for example only the parameters `pulse_start`, `pulse_length` and `period` and let the template do calculations internally)

qtt/instrument_drivers/virtualAwg/templates.py
```diff
@@ -60,3 +60,19 @@ def marker(name):
...
+
+        Args:
+            name (str): The user defined name of the sequence.
+
+        Returns:
+            TablePT: The sequence with the wait pulse and rollover.
```
eendebakpt:
```suggestion
            TablePT: The sequence with the marker pulse and rollover part of the pulse.
```

PR #515, opened 2019-03-07
[DEM-665] Ripple in sawtooth in videomode with HDAWG/UHFLI

qSaevar:


CumulonimbusCalvus:
tmp 

I still see the multiple sawteeth?

PR #518, opened 2019-03-12
Added code coverage

CumulonimbusCalvus:
@peendebak 

- I added the coverage. Other analysis tools require some investigation, because they are not always free (>4 developers).

- I removed the pip cache statement. We would like to see how qtt installs from scratch without caching old pip-package installations.

- The current PyPi QCoDeS requires numpy < 1.14 where we also have Polygon3 raising errors when numpy is <1.14... How to solve this in a robust manner? Installing QCoDeS directly from the git master maybe?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: python
-cache: pip
 
 matrix:
   include:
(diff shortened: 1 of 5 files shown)


peendebak:
@CumulonimbusCalvus Polygon3 works fine on numpy 1.15 for me. But the package seems to be unmaintained, to it would be good to find a replacement. I could not find anything quickly, so this might be some work.

peendebak:
@CumulonimbusCalvus I think we want to move the tests from `qtt/qtt/tests` to `qtt/tests` , otherwise the tests will fall under the coverage.

PR #520, opened 2019-03-13
Move UnitTests

CumulonimbusCalvus:
@peendebak 

I moved the unittests in this PR as requested [here](https://github.com/QuTech-Delft/qtt/pull/518#issuecomment-472173446). I have taken the structure from the demonstator repo's to have consistency.

I still should check whether readthedocs is still working and other additional stuff besides installing, running en unittesting.


.travis.yml
```diff
@@ -20,12 +20,12 @@ install:
...
         pip install -r develop_requirements.txt;
       fi
-    - python setup.py develop
+    - pip install -e .
 
 script:
     - ls
     - pytest --version
-    - coverage run --source="./qtt" --omit="*/test_*"  -m pytest
+    - coverage run --source="./src" --omit="*/test_*"  -m pytest
```
peendebak: I guess we can remove the omit part now?
```suggestion
    - coverage run --source="./src"  -m pytest
```

CumulonimbusCalvus: Already done.

CumulonimbusCalvus:
@peendebak 

All OK I think. There is one file (src/qtt/exampledata/misc/P5_y_P5_x.dat) which is changed according to the PR. Still, when I check for diff's there is nothing changed.

peendebak:
@CumulonimbusCalvus Merged. I create a new PR for additional cleanup of the test structure.

PR #521, opened 2019-03-15
Updates from XLD fridge

peendebak:
* Various small fixes
* Remove legacy code in `sensingdot_t`
* Fix bug in `get_dataset`

### DIFF
--- a/qtt/simulation/dotsystem.py
+++ b/qtt/simulation/dotsystem.py
@@ -15,7 +15,7 @@ import logging
 
 try:
     import graphviz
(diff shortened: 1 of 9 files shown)


qtt/simulation/dotsystem.py
```diff
@@ -15,7 +15,7 @@
 
 try:
     import graphviz
-except:
+except BaseException:
```
CumulonimbusCalvus:
```suggestion
except ModuleNotFoundError:
```

PR #556, opened 2019-04-17
[DEM-890] implement signal processor and signal processor interface

ammar92:


src/qtt/measurements/new/signal_processor.py
```diff
@@ -0,0 +1,33 @@
...
+        self._signal_processors: List[SignalProcessorInterface] = []
+
+    def add_signal_processor(self, signal_processor: SignalProcessorInterface) -> None:
+        """ Add a signal processor.
+
+            Args:
+                signal_processor: A signal processor
+        """
+
+        self._signal_processors.append(signal_processor)
```
quantumkoen: Do we need a run-time check that what is passed in is indeed an implementation of `SignalProcessorInterface` and raise a `TypeError` otherwise?

src/qtt/measurements/new/signal_processor.py
```diff
@@ -0,0 +1,33 @@
+from typing import List
+
+from qilib.data_set import DataSet
+
+from qtt.measurements.new.interfaces.signal_processor_interface import SignalProcessorInterface
+
+
+class SignalProcessor:
```
quantumkoen: I think the naming here is confusing. We have `SignalProcessorInterface`, and `SignalProcessor`. At first sight I expect a `SignalProcessor` to be an implementation of `SignalProcessorInterface`, but this is not the case. I would suggest renaming `SignalProcessor` to `SignalProcessorRunner` or something like that, so that the name actually describes what it is and avoids confusion.

src/qtt/measurements/new/signal_processor.py
```diff
@@ -0,0 +1,33 @@
...
+    def add_signal_processor(self, signal_processor: SignalProcessorInterface) -> None:
+        """ Add a signal processor.
+
+            Args:
+                signal_processor: A signal processor
+        """
+
+        self._signal_processors.append(signal_processor)
+
+    def run_processes(self, signal_data: DataSet) -> DataSet:
```
quantumkoen: Again, naming could be better: `run_processes` suggest we fork things and create processes.. Which is not the case. Simply `run` would probably be enough, or if you want to be verbose `run_processors`, which is more accurate.

src/tests/measurements/new/test_signal_processor.py
```diff
@@ -0,0 +1,48 @@
...
+        class DummySignalProcessor(SignalProcessorInterface):
+            def run_process(self, signal_data: DataSet) -> DataSet:
+                return signal_data
+
+        signal_processor = SignalProcessor()
+        signal_processor.add_signal_processor(DummySignalProcessor())
+        self.assertEqual(len(signal_processor._signal_processors), 1)
+        self.assertIsInstance(signal_processor._signal_processors[0], DummySignalProcessor)
+
+    def test_run_process_single_signal_processor(self):
```
quantumkoen: I'd expect a test-case for at least running it with two processors as well, and check that the final result is actually processor1(processor2(input). And perhaps running with 0 as well, just to cover the boundary cases. What is the expected behaviour of running with 0?

src/tests/measurements/new/test_signal_processor.py
```diff
@@ -0,0 +1,48 @@
...
+from numpy import array_equal
+from numpy.ma import array
+from qilib.data_set import DataSet, DataArray
+
+from qtt.measurements.new.interfaces.signal_processor_interface import SignalProcessorInterface
+from qtt.measurements.new.signal_processor import SignalProcessor
+
+
+class TestSignalProcessor(TestCase):
+    def test_run_processes(self):
```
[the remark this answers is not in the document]
ammar92: Done

src/tests/measurements/new/test_signal_processor.py
```diff
@@ -0,0 +1,48 @@
...
+        with patch('qtt.measurements.new.interfaces.signal_processor_interface.SignalProcessorInterface') as spi:
+            class DummySignalProcessor(spi):
+                def run_process(self, signal_data: DataSet) -> DataSet:
```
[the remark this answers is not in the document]
ammar92: Done

src/qtt/measurements/new/signal_processor.py
ammar92: Agree. @CumulonimbusCalvus what do you think?

src/qtt/measurements/new/signal_processor.py
ammar92: Done

src/tests/measurements/new/test_signal_processor.py
ammar92: Good one. Done

src/qtt/measurements/new/signal_processor_runner.py
```diff
@@ -0,0 +1,33 @@
...
+        self._signal_processors: List[SignalProcessorInterface] = []
+
+    def add_signal_processor(self, signal_processor: SignalProcessorInterface) -> None:
+        """ Add a signal processor.
+
+            Args:
+                signal_processor: A signal processor
+        """
+
+        self._signal_processors.append(signal_processor)
```
quantumkoen: Repeating an earlier comment: Do we need a run-time check that what is passed in is indeed an implementation of SignalProcessorInterface and raise a TypeError otherwise?

ammar92: Done

src/qtt/measurements/new/signal_processor_runner.py
```diff
@@ -18,6 +18,9 @@ def add_signal_processor(self, signal_processor: SignalProcessorInterface) -> No
                 signal_processor: A signal processor
         """
 
+        if not isinstance(signal_processor, SignalProcessorInterface):
+            raise TypeError('signal_processor must be of type SignalProcesorInterface')
```
quantumkoen: You should add a testcase to cover this exception.

PR #562, opened 2019-05-01
[DEM-928] As an operator i want to measure according to Elzerman readout

qSaevar:
🌴

### DIFF
--- a/docs/notebooks/unfinished/example_measure_elzerman_readout.py
+++ b/docs/notebooks/unfinished/example_measure_elzerman_readout.py
@@ -5,16 +5,19 @@ from matplotlib import pyplot as plt
 # INITIALIZE THE SCOPE READER
 
 device_id = 'dev2338'
-time_period = 2
+time_period = 3
 trigger_enabled = True
 input_channel = 1
 input_channel_attribute = 'Demod 1 R'
+nearest_sample_rate = 27e3
+
 scope_reader = UHFLIScopeReader(address=device_id)
-sampling_rate = scope_reader.get_nearest_sample_rate(27e3)
 scope_reader.period = time_period
-scope_reader.trigger_enabled = trigger_enabled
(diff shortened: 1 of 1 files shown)


docs/notebooks/unfinished/example_measure_elzerman_readout.py
```diff
@@ -0,0 +1,54 @@
+from qtt.measurements.acquisition import UHFLIScopeReader, UHFLIStimulus
```
peendebak: Rename the file to something else, the example is about triggered readout of the device, but not related to Elzerman.

docs/notebooks/unfinished/example_measure_elzerman_readout.py
```diff
@@ -0,0 +1,54 @@
...
+
+# INITIALIZE THE SCOPE READER
+
+device_id = 'dev2338'
+time_period = 2
+trigger_enabled = True
+input_channel = 1
+input_channel_attribute = 'Demod 1 R'
+scope_reader = UHFLIScopeReader(address=device_id)
+sampling_rate = scope_reader.get_nearest_sample_rate(27e3)
```
peendebak: The `sampling_rate` is not used in the script. Should it be set on the `UHFLIScopeReader`?

docs/notebooks/unfinished/example_measure_elzerman_readout.py
peendebak: This example script cannot be executed without the UHFLI hardware. If this is converted to a jupyter notebook, then the output that is generated by the script can also be included. E.g. the output of `print(records)`

Would it be an option to do that?

CumulonimbusCalvus: Maybe we can add the Elzerman related stuff next week and turn it into a full sample? We could then also convert the sample into a notebook.

PR #563, opened 2019-05-02
[DEM-888] 2D post processing

CumulonimbusCalvus:
🌤


src/qtt/measurements/post_processing/process_sawtooth_2d.py
```diff
@@ -0,0 +1,67 @@
...
+
+        Returns:
+            A data set which contains a 2D image with the charge stability diagram.
+        """
+        width_x, width_y = data_set.user_data['width']
+        resolution_x, resolution_y = data_set.user_data['resolution']
+
+        output_data_set = DataSet()
+        output_data_set.user_data = data_set.user_data
+        for _, data_array in data_set.data_arrays.items():
```
ammar92: Because you're not using the keys, then maybe it's better to iterate over the dictionary values rather than items? E.g. 
```python
for data_array in iter(data_set.data_arrays.values())
```

src/qtt/measurements/post_processing/process_sawtooth_2d.py
```diff
@@ -0,0 +1,67 @@
...
+                      user data of the data set should contain the width and resolution settings.
+
+        Returns:
+            A data set which contains a 2D image with the charge stability diagram.
+        """
+        width_x, width_y = data_set.user_data['width']
+        resolution_x, resolution_y = data_set.user_data['resolution']
+
+        output_data_set = DataSet()
+        output_data_set.user_data = data_set.user_data
```
ammar92: I would personally put this in the DataSet's constructor like this:
```python
output_data_set = DataSet(user_data=data.user_data)
```
However, this is fine and readable as well. I'll leave this up to you.

PR #576, opened 2019-05-26
deprecate legacy methods

peendebak:


src/qtt/data.py
```diff
@@ -1068,7 +1068,7 @@ def experimentFile(outputdir: str = '', tag=None, dstr=None, bname=None):
     pfile = os.path.join(outputdir, tag, basename + '.' + ext)
     return pfile
 
-
+@qtt.utilities.tools.rdeprecated(expire='1-1-2020')
```
QFer: rdeprecated has a parameter 'txt' to help the user of this deprecated function to find the alternative or explaining the reason that this function is deprectated. We should use this parameter for deprecated functions.

src/qtt/data.py
```diff
@@ -1078,6 +1078,7 @@ def loadExperimentData(outputdir, tag, dstr):
     return dataset
 
 
+@qtt.utilities.tools.rdeprecated(expire='1-1-2020')
```
QFer: idem. parameter txt should be used.

peendebak:
> What does this expire date mean in qtt? When I do a search over the code I see 17 instances of deprecated methods with expired deprecation date but they still exist.
> 
> The commit seems out of date due to another commit earlier. Rebase is needed.

Before the specified data the decorator will issue a warning, after the specified data it will throw an error. @QFer

PR #578, opened 2019-05-27
[OP-32] Add python37 to travis

peendebak:


QFer:
@CumulonimbusCalvus @peendebak OP-32 is ready for review.
* Coverage is now 43.8%. Because of wrong usage, coverage didn't omit the files to omit. This has been repaired. Also, because of OP-34 coverage is now only counting real production code in qtt/src and tests are found in qtt/tests only. So we are not counting testcode anymore in coverage.

PR #591, opened 2019-06-18
[OP-48] Added a changelog to QTT.

CumulonimbusCalvus:
* Added the change-log to the root folder.
* Note that the **Codecy** currently fails. This is because the CHANGELOG.md is checked. The markdown file can only be removed form the **Codecy** checks once it is in the dev-branch.


docs/changes/1.1.2.rts
```diff
@@ -0,0 +1,18 @@
+Changelog for QTT 1.1.2
+==========================
+
+This release changes the folder structure and starts seperating the unittests from the source code.
```
QFer: Actually we're finished seperating unit tests from the source code. pytest is also not looking for tests in the source directory anymore (see pytest.ini). Further Coverage is only counting code coverage in directory src/qtt, not the src/tests directory (see .travis.yml)

docs/changes/1.1.2.rts
```diff
@@ -0,0 +1,18 @@
...
+
+
+New & Improved
+______________
+
+- Last version before adding a changelog.
+- Source code is now located in the ./qtt/src/ folder 
+- Most of the unittests are now located in the ./qtt/src/tests/ folder.
```
QFer: All of the unittests

docs/changes/1.1.2.rts
```diff
@@ -0,0 +1,18 @@
+Changelog for QTT 1.1.2
```
QFer: I think each version should have a release date.

CumulonimbusCalvus:
> 
> 
>     * I think the changelog should be in the root directory where you can find it immediately, not somewhere hidden in docs\changes.
> 
>     * We should have exactly one changelog with all versions in it, not 1 file for each version. Otherwise we have to look in multiple files. So a /changes/ directory to keep the files together should not be needed.
> 
>     * With one changelog instead of a log for each version, we can rename the changelog to CHANGELOG.md
> 
>     * We should describe what's added, changed, bug fixed, deprecated. Maybe you should take a look at: https://keepachangelog.com/en/1.0.0/ for a nice example and some more hints about changelogs.

Nice suggestions. I will implemented them. We can also use the semantic versioning as discussed. The long term support (LTS) idea outcome of the software meeting can then be put into MAJOR updates.

CHANGELOG.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Fixed
+- Fixed the readthedocs build (#590).
+- Fixed issue with HDAWG8 sampling rate (#589).
+- Fixed travis warnings and tests instrument_drivers (#569).
+
+### Security
+...
+
+## [1.1.2] - 2018-04-12
```
QFer: Date earlier than previous version?

CumulonimbusCalvus:
```suggestion
## [1.1.2] - 2019-04-12
```

PR #594, opened 2019-06-25
[OP-51] Videomode refactor

peendebak:
* Add abstract base class `VideoModeProcessor`
* Refactor to workflow:

        i) Initialize. For example start a periodic waveform on the AWG
        ii) Start the readout. This starts a loop with the following steps:
            - Measure data
            - Post-process data
            - Plot data
            The loop continues running in the background untill the user aborts the loop.
        iii) Stop the readout. This stops the measure-process-plot loop
        iv) Stop. This stops all activity (e.g. both the readout loop and and activity on the AWG)

* Move ttrace tests into proper module
* Add unit tests for `MeasurementControl` and `livePlot`
* Fix issue with importing M4i if it is not installed
* Update example notebook


peendebak:
@CumulonimbusCalvus Checked with the Keysight/M4i on XLD fridge.

PR #597, opened 2019-06-27
[OP-52] Add boxcar filter.

mjwoudstra:


src/qtt/algorithms/generic.py
```diff
@@ -581,10 +585,47 @@ def weightedCentroid(im, contours, contourIdx, fig=None):
...
+
+def boxcar_filter(signal, kernel_size):
+    """ Perform boxcar filtering on an array.
+    At the edges, the edge value is replicated beyond the edge as needed by the size of the kernel.
+    This is the 'nearest' mode of scipy.ndimage.convolve. For details, see
+    https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.convolve.html?highlight=mode
+
+
+    Args:
+        signal (sequence): a sequence containing the signal to be filtered.
```
peendebak:
```suggestion
        signal (array): An array containing the signal to be filtered.
```
The term sequence already as other meanings. Also I would not call a 2D array a 'sequence'

src/qtt/algorithms/generic.py
```diff
@@ -581,10 +585,47 @@ def weightedCentroid(im, contours, contourIdx, fig=None):
...
+    """
+
+    if not isinstance(signal, np.ndarray):
+        signal = np.array(signal)
+
+    if len(kernel_size) != len(signal.shape):
+        raise RuntimeError('Number of dimensions of kernel (%d) not equal to number of dimension of input signal (%d)' %
+                           (len(kernel_size), len(signal.shape)))
+
+    for siz in kernel_size:
```
peendebak: Replace the next two lines with:
```
if np.any(kernel_size<=0):
   raise ...
```

src/qtt/algorithms/generic.py
```diff
@@ -581,10 +585,47 @@ def weightedCentroid(im, contours, contourIdx, fig=None):
...
+    boxcar_kernel = np.ones(kernel_size, dtype=np.float64) / np.prod(kernel_size, dtype=np.float64)
+    filtered_signal = scipy.ndimage.filters.convolve(filtered_signal, boxcar_kernel, mode='nearest')
+
+    return filtered_signal
+
+
 # %%
 
 
 if __name__ == '__main__':
```
peendebak: Remove these two lines, they do not do anything

src/tests/algorithms/test_generic.py
```diff
@@ -40,3 +40,169 @@ def test_rescale_image(self):
...
+        with self.subTest(signal=signal):
+            out_fnd = boxcar_filter(signal, kernel)
+            msg = ''
+            try:
+                np.testing.assert_allclose(out_fnd, out_exp)
+            except AssertionError as e:
+                msg = '\nsignal: %s\ngives wrong filtered output:' % (signal,)
+                msg += str(e).replace('x:', '  actual:').replace('y:', 'expected:')
+
+            if msg:
```
peendebak: You can remove the `if msg:` by moving the statement into the `except` part.

src/tests/algorithms/test_generic.py
```diff
@@ -40,3 +40,169 @@ def test_rescale_image(self):
...
+                np.testing.assert_allclose(int_signal, signal)
+            except AssertionError:
+                continue
+
+            self.assertFilterOutput(int_signal, kernel_1D, out_exp)
+
+        kernel_2D = TestBoxcarFilter._kernel_size_2D
+        for signal, out_exp in TestBoxcarFilter._known_inputs_outputs_2D:
+            int_signal = np.array(signal, dtype=int)
+            try:
```
peendebak: Replace the `try-except` block with
```
with self.assertRaises(AssertionError):
   np.testing.assert_allclose(int_signal, signal)
```
(similar for some other methods)

src/tests/algorithms/test_generic.py
mjwoudstra: That is true from the program logic point of view and that is what it was initially. However, it turns out that if you do that, you get a warning that an exception was raised (by self.fail()) inside an exception handler. It does not seem to like that.

src/qtt/algorithms/generic.py
mjwoudstra: Fine with me.

src/qtt/algorithms/generic.py
mjwoudstra: Ah, those fancy numpy functions

src/qtt/algorithms/generic.py
mjwoudstra: Agreed!

src/tests/algorithms/test_generic.py
mjwoudstra: The goal of the code is to only process those arrays whose float values are actually integers. The suggested replacement does not replicate that behavior. I have replaced the `np.testing.assert_allclose() + try/except` 
with 
`if np.allclose():`

src/qtt/algorithms/generic.py
mjwoudstra: Just need to convert the kernel_size to a np.array first

peendebak:
@mjwoudstra Fixed some more issues with the new qcodes version tag and rebased to dev.

peendebak:
@CumulonimbusCalvus Fixes issues with the new mypy and the new version tag of qcodes. Can you review that part?

mjwoudstra:
> 
> 
> @mjwoudstra Fixed some more issues with the new qcodes version tag and rebased to dev.

@peendebak  That rebase also wiped out my TestBoxcarFilter test class. I put it back in commit 3e2b038.
In commit 2d6831c   I implemented a workaround for the packaging dependency suggested on qcodes slack by jenshnielsen this morning:
https://qcodes.slack.com/archives/C18KR2HD3/p1563953276010500

CumulonimbusCalvus:
@mjwoudstra @peendebak 

I moved the solution for #604 to PR #611, since this problems has nothing to do with adding a boxcar filter. Once done I will S&M, rebase and revert the updates related to #604.

CumulonimbusCalvus:
@mjwoudstra Should I rebase this branch or can you do it?

mjwoudstra:
@CumulonimbusCalvus I will rebase it to dev.

peendebak:
@mjwoudstra Can you add a single line to https://github.com/QuTech-Delft/qtt/blob/dev/CHANGELOG.md in this PR? 

I think we can merge then.

PR #605, opened 2019-07-24
[DEM-1018] VirtualDacInstrumentAdapter needs to set instead of append instruments

ammar92:


src/qtt/instrument_drivers/gates.py
```diff
@@ -93,7 +96,19 @@ def gate_map(self):
     @property
     def instruments(self):
         """ A list of QCoDeS instruments."""
-        return self._instrument_list
+        return self._instruments
+
+    @instruments.setter
+    def instruments(self, value: List[Instrument]):
```
peendebak: Updating the `instruments` without updating the `gate_map` creates situations that are not safe. Updating the instruments after the `VirtualDAC` has been created should only be done in a method that also updates the `gate_map`.

src/qtt/instrument_drivers/gates.py
```diff
@@ -175,24 +190,31 @@ def _remove_gates(self):
             self.functions.pop(f'get_{gate}')
             self.functions.pop(f'set_{gate}')
 
-    def add_instruments(self, instruments):
+    def add_instruments(self, instruments: List[Instrument]):
```
peendebak: Policy is to either use type annotations in the entire file, or not at all.

src/qtt/instrument_drivers/gates.py
ammar92: Using qilib's `VirtualDACInstrumentAdapter` the `gate_map` is updated properly after adding the instruments. Without using the instrument adapter, I think it should be the user's responsibility to update `gate_map`, as there is also a setter for this. The `instruments` setter and `add_instruments` is just a convenient way to update the instrument's list.

src/qtt/instrument_drivers/gates.py
ammar92: Removed for now, but where can I find this policy?

src/qtt/instrument_drivers/gates.py
peendebak: @ammar92 The instrument adapters are not compatible with the standard qcodes initialization methods, so most people will not be using them. If you want to update the instruments, then remove the gate map (set it to an empty dict)  and the user can add it later. If only the instruments are updated, then the instruments and gate_map do not match and that is something we do not want.

ammar92: You're right. This is hopefully now resolved my invalidating/emptying `gate_map` when updating the instruments.

PR #613, opened 2019-07-31
[OP-61] Improve qtt installation.

peendebak:
* update installation and test instructions
* remove unnecessary extra modules

@mjwoudstra @CumulonimbusCalvus

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -70,7 +70,7 @@ setup(name='qtt',
           'numpy>=1.15', 'scikit-image', 'IPython>=0.1', 'qcodes>=0.4', 'Polygon3',
           'scipy', 'pyqtgraph', 'qupulse'
       ],
-      tests_require=['unittest'],
+      tests_require=['pytest'],
       extras_require=extras_require,
       zip_safe=False,
       )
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -70,10 +66,11 @@ def get_version(verbose=1, filename='src/qtt/version.py'):
       package_dir={'': 'src'},
       packages=find_packages(where='./src'),
       install_requires=[
-          'matplotlib', 'pandas', 'attrs', 'dulwich', 'qtpy', 'nose', 'hickle', 'pyzmqrpc',
+          'matplotlib>=3.0', 'pandas', 'attrs', 'dulwich', 'qtpy', 'nose', 'hickle', 'pyzmqrpc',
           'numpy>=1.15', 'scikit-image', 'IPython>=0.1', 'qcodes>=0.4', 'Polygon3',
           'scipy', 'pyqtgraph', 'qupulse'
       ],
+      tests_require=['unittest'],
```
CumulonimbusCalvus:
```suggestion
      tests_require=['pytest'],
```

The module unittest comes with python preinstalled, pytest on the other hand not.

mjwoudstra:
pyqtgraph needs PyQt5 and the installation with `pip install -e .`  fails if PyQt5 is not installed. This can be solved by adding pyqt5 to the list of `install_requires`

Issue #610, opened 2019-07-31
unittests fail with ModuleNotFoundError: No module named 'mock'


mjwoudstra:
NB: if you have the external module 'mock' installed, you will probably not see this error.

Running pytest on the dev branch in a fresh environment gives the following import errors:
`___________________________________________________________ ERROR collecting src/tests/unittests/measurements/acquisition/test_uhfli_stimulus.py ____________________________________________________________
ImportError while importing test module '<workdir>\qtt\src\tests\unittests\measurements\acquisition\test_uhfli_stimulus.py'.
Traceback:
src\tests\unittests\measurements\acquisition\test_uhfli_stimulus.py:3: in <module>
    from mock import patch, MagicMock, call
E   ModuleNotFoundError: No module named 'mock'
_______________________________________________________ ERROR collecting src/tests/unittests/measurements/processing/test_signal_processor_runner.py ________________________________________________________
ImportError while importing test module '<workdir>\qtt\src\tests\unittests\measurements\processing\test_signal_processor_runner.py'.
Traceback:
src\tests\unittests\measurements\processing\test_signal_processor_runner.py:3: in <module>
    from mock import patch
E   ModuleNotFoundError: No module named 'mock'`

The solution is described here:
https://stackoverflow.com/questions/11501520/importerror-no-module-named-mock
It seems that mock is now part of unittest, so:
`from unittest.mock import bla`
should work out-of-the-box and should not require the external module mock (I have not tried that though).

CumulonimbusCalvus:
This issue has not been resolved by #613. Will create a new PR which removes all the `from mock ...` imports.

CumulonimbusCalvus:
@mjwoudstra 

I removed all the `from mock import ...` in PR #620. Does this solve your issue?

PR #620, opened 2019-08-02
[OP-64] Move mock imports to unittest.mock.

CumulonimbusCalvus:
Currently there are unittests which import mocks like `from mock import ...`. The mocking framework is also available directly via `unittest.mock`.

Changed all mock import to use the `unittest.mock`.
Should solve issue #610.


mjwoudstra:
There seems to be a typo:  unitest (one t) instead of unittest (2 t's). See Travis build log.

PR #624, opened 2019-08-08
[DEM-1045] Fix qilib for readthedocs build

oschusler:


requirements.txt
```diff
@@ -16,5 +16,4 @@ sympy
 scikit-image
 scipy>=0.18
 tables
-git+https://github.com/QuTech-Delft/qilib.git#egg=qilib
-qcodes
\ No newline at end of file
+qcodes
```
peendebak:
```suggestion
qcodes
qilib
```
We need the packages in both `setup.py` and the `requirements.txt`

oschusler: fixed

PR #627, opened 2019-08-15
[OP-66] autodetect notebooks

QFer:
* Travis test shell script to find the notebooks to run
* Excludes directory measurements and unfinished


.travis.yml
```diff
@@ -49,7 +49,9 @@ script:
     - mypy src --ignore-missing-imports
     - if [[ "$FULL_INSTALL" == "1" ]]; then
         jupyter --version;
-        source docs/notebooks/run_notebooks.sh;
+        chmod +x docs/notebooks/run_notebooks.sh;
```
peendebak: Rename `run_notebooks.sh` do something more descriptive. (e.g. `select_notebooks_to_execute`)

docs/notebooks/run_notebooks.sh
```diff
@@ -1,20 +1,19 @@
...
+#!/bin/bash
+#Collect all notebooks (except the ones in the measurements and unfinished folder) to be executed
+#One or more notebooks from the measurements folder or others can be added manually by copying, uncommenting
+#and adjusting the filename of the following line and appending it to the bottom of this script.
+#echo "jupyter nbconvert --to notebook --execute docs/notebooks/measurements/some_notebook_to_add.ipynb" >>$OUTFILE
 
+echo "Collecting notebooks to run"
+OUTFILE="docs/notebooks/run_selected_notebooks.sh"
+/bin/cat <<EOM >$OUTFILE
+# Generated file, do not edit it or put it in github
```
peendebak: Maybe add the generated file to `.gitignore`

.travis.yml
```diff
@@ -49,7 +49,9 @@ script:
...
+        bash docs/notebooks/run_notebooks.sh;
```
peendebak: Is there a reason for generating the `run_selected_notebooks.sh` instead of executing the commands directly?

.travis.yml
QFer: I will refactor the filename.

docs/notebooks/run_notebooks.sh
QFer: I will

.travis.yml
QFer: No reason anymore with the current approach. I will change it so we don't need a temporary file. It didn't work with an earlier version.

docs/notebooks/run_notebooks.sh
QFer: because we don't use a generated file anymore, .gitignore need not be adjusted anymore.

peendebak:
@QFer Thanks!

PR #631, opened 2019-08-19
[OP-72] Add method to restrict boundaries on VirtualDAC

peendebak:
* Add restrict method
* Change `__repr__` for `VirtualDAC`


src/qtt/instrument_drivers/gates.py
```diff
@@ -246,9 +246,22 @@ def get_boundaries(self):
             boundaries[gate] = param.vals.valid_values
         return boundaries
 
+    def restrict_boundaries(self, gate_boundaries):
+        """ Restrict boundaries values that can be set on the gates.
+
+        Args:
+            gate_boundaries (dict): a range of allowed values per parameter.
+        """
+        new_boundaries = self.get_boundaries()
```
QFer: new_boundaries should be renamed to current_boundaries
to get rid of the usage before assignment in the last line of this method you should check first if number of gate_boundaries > 0, then the new_boundaries in the for-loop is assigned to for sure. Plus you don't reassign when there are no gate_boundaries given.

src/qtt/instrument_drivers/gates.py
```diff
@@ -246,9 +246,22 @@ def get_boundaries(self):
...
+
+        for g, bnds in gate_boundaries.items():
```
QFer: g should be renamed to gate, bnds to boundary. It's all in the name, keep it as clear as possible.

src/qtt/instrument_drivers/gates.py
```diff
@@ -246,9 +246,22 @@ def get_boundaries(self):
             boundaries[gate] = param.vals.valid_values
         return boundaries
 
+    def restrict_boundaries(self, gate_boundaries):
+        """ Restrict boundaries values that can be set on the gates.
+
+        Args:
+            gate_boundaries (dict): a range of allowed values per parameter.
```
QFer: gate_boundaries is not the range that will be allowed. It must fit within the current boundary of that gate. This docstring fits the method set_boundaries, not restrict_boundaries

src/tests/unittests/instrument_drivers/test_gates.py
```diff
@@ -66,6 +66,15 @@ def test_get_boundaries(self):
 
         virtual_dac.close()
 
+    def test_restrict_boundaries(self):
+        gate_map = {'P1': (0, 1), 'P2': (0, 2), 'P3': (0, 3)}
+        virtual_dac = VirtualDAC(instrumentName('test'), instruments=[self.ivvi], gate_map=gate_map)
+        gate_boundaries = {'P1': (0, 100), 'P2': (0, 100)}
+        virtual_dac.set_boundaries(gate_boundaries)
+        virtual_dac.restrict_boundaries({'P1': (-50, 50), 'P3': (0, 1)})
```
QFer: I tried virtual_dac.restrict_boundaries(({'P1': (150, -150)}) here. Do you know what happens then? Is this as expected? In real OO boundary would be a class instead of a dictionary. Within this class relations between the values can be checked on the fly when being set in the constructor for example.

QFer: Also tried virtual_dac.restrict_boundaries({}) here. There happens more than I expected.

peendebak: I added a check to make sure a boundary (a,b) satisfies a<b.

PR #632, opened 2019-08-20
fixing a bug in the function setSingleStep by adding a try statement

NicoHendrickx:
The for-loop cycles through all objects in the parameter viewer dict. The first one of each instrument is the QTreeWidget, where it expects a dict. Adding the try-except solves the issue.

@peendebak


peendebak:
@NicoHendrickx Closing this. I created #633 which has a similar solution but with unit tests.

PR #633, opened 2019-08-20
fix single step method of parameterviewer

peendebak:
@NicoHendrickx


NicoHendrickx:
Why do you specifically check for a _treewidgetitem to continue? Isn't the inverse (checking if it's a valid type, instead of checking for 1 invalid one) better, or keeping the try/except as I suggested? In case later different type items get added to the viewer.

peendebak:
> Why do you specifically check for a _treewidgetitem to continue? Isn't the inverse (checking if it's a valid type, instead of checking for 1 invalid one) better, or keeping the try/except as I suggested? In case later different type items get added to the viewer.
The `try-except` from the other PR silences all errors, which we do not want. I agree that the specific check on `_treewidgetitem` is not really nice.

PR #640, opened 2019-08-27
[OP-79] Add automatic low-pass filter to Coulomb peak analysis

peendebak:
* add low-pass filtering code
* add automatic filtering to Coulomb peak analysis
* Add example notebook for low-pass filtering


peendebak:
Replaced by #641

PR #655, opened 2019-09-17
add option to set dataset record labels through scanjob

peendebak:


src/qtt/measurements/scans.py
```diff
@@ -349,6 +349,10 @@ def _initialize_live_plotting(alldata, plotparam, liveplotwindow=None, subplots=
         pyqtgraph.mkQApp().processEvents()  # needed for the parameterviewer
     return liveplotwindow
 
+def _dataset_record_label(scanjob):
+    """ Return label to be used in dataset record """
```
CumulonimbusCalvus: The argument `scanjob` en return value are not documented yet.

src/qtt/structures.py
```diff
@@ -467,7 +467,7 @@ def fastTune(self, Naverage=90, sweeprange=79, period=.5e-3, location=None,
             scanjob['minstrument'] = channel
             scanjob['minstrumenthandle'] = instrument
             scanjob['wait_time_startscan'] = sleeptime
-
+            scanjob['dataset_label'] = 'sensingdot_fastune'
```
CumulonimbusCalvus:
```suggestion
            scanjob['dataset_label'] = 'sensingdot_fast_tune'
```

Is this a typo?

src/tests/unittests/measurements/test_scans.py
```diff
@@ -74,6 +74,26 @@ def test_scan1D(self, verbose=0):
         self.assertTrue('hi' in data.metadata)
         gates.close()
 
+    @staticmethod
+    def test_scan1D_no_gates():
+        p = Parameter('p', set_cmd=None)
+        r = VoltageDivider(p, 4)
+        scanjob = scanjob_t({'sweepdata': {'param': p, 'start': 0, 'end': 10, 'step': 2}, 'minstrument': [r]})
+        station=qcodes.Station()
+        _ = scan1D(station, scanjob, liveplotwindow=False, verbose=0)
```
CumulonimbusCalvus: This test does not check anything. It would be nice to test whether the 'label' value is indeed what you expect for the default case, such that the added function `_dataset_record_label` works fully as expected.

src/qtt/measurements/scans.py
peendebak: > Some minor changes requested. In the future it would be nice to change the `scanjob` into a class. Then the `dataset_record_label`, etc. can become a property of `scanjob`.

Yes i agree. The reason it is not yet a a class is that serialization is an issue then.

peendebak: Method was private. I will add docs anyway

PR #663, opened 2019-09-26
[OP-86] Add example notebook for serialization

peendebak:
* Add example notebook for serialization
* Add methods for serialization of dataclasses

### DIFF
--- a/docs/examplenotebook.rst
+++ b/docs/examplenotebook.rst
@@ -24,6 +24,17 @@ Analysis
 
     notebooks/analysis/example_ohmic.ipynb
     notebooks/analysis/*
+
+Data
+----
+
+.. toctree::
+    :maxdepth: 3
+    :titlesonly:
+    :glob:
+
+    notebooks/datasets/*
+
     
 Simulation
 ----------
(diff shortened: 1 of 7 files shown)


peendebak:
> @peendebak
> 
> The addition of the encode and decode json_dataclasses is very generic. I think this addition would be more helpful if it would be added to qilib. Schould I do it?
@CumulonimbusCalvus If you could add it to qilib: great! You might consider renaming the functions, since they will work for any class that has a `.to_dict()` and a class method `.from_dict()`.

The notebook should stay in `qtt`, unless you want to setup the infrastructure to publish notebooks to readthedocs also for `qilib`.

CumulonimbusCalvus:
The encode and decode json_dataclasses have been added to QiLib, see [PR-58](https://github.com/QuTech-Delft/qilib/pull/58/files) in qilib.

After S&Q and a new release on PyPi we can S&Q this PR.

CumulonimbusCalvus:
@peendebak I removed the dataclass serialization from QTT and added it to QILib. We can S&M this PR once a new release has been made of QILib with the addition.

see [PR-58](https://github.com/QuTech-Delft/qilib/pull/58/files)

PR #664, opened 2019-09-27
[OP-95] Add the AllXY analysis code.

peendebak:


src/tests/unittests/algorithms/test_allxy.py
```diff
@@ -0,0 +1,40 @@
+import numpy as np
+import unittest
+import matplotlib.pyplot as plt
+
+import qtt
```
CumulonimbusCalvus:
```suggestion
from qtt.data import makeDataSet1Dplain
```

src/tests/unittests/algorithms/test_allxy.py
```diff
@@ -0,0 +1,40 @@
...
+from qtt.algorithms.allxy import fit_allxy, plot_allxy, allxy_model
+
+
+class TestAllxy(unittest.TestCase):
+
+    def test_allxy(self):
+        dataset = qtt.data.makeDataSet1Dplain('index', np.arange(21), 'allxy', [0.12, 0.16533333, 0.136, 0.17066666, 0.20266667,
```
CumulonimbusCalvus: ### 
```suggestion
        dataset = makeDataSet1Dplain('index', np.arange(21), 'allxy', [0.12, 0.16533333, 0.136, 0.17066666, 0.20266667,
```

src/qtt/algorithms/allxy.py
```diff
@@ -0,0 +1,128 @@
...
+    """ Generate all combinations of the AllXY sequence from Reed 2013 """
+    xymapping = {'I': 'I', 'x': 'X90', 'y': 'Y90', 'X': 'X180', 'Y': 'Y180'}
+    allxy_combinations_input = ['II', 'XX', 'YY', 'XY', 'YX'] + ['xI', 'yI', 'xy', 'yx',
+                                                                 'xY', 'yX', 'Xy', 'Yx', 'xX', 'Xx', 'yY', 'Yy'] + ['XI', 'YI', 'xx', 'yy']
+    allxy_combinations = [(xymapping[x[0]], xymapping[x[1]]) for x in allxy_combinations_input]
+
+    return allxy_combinations
+
+
+def allxy_model(x: Union[float, np.ndarray], offset0: float, slope0: float, offset1: float, slope1: float, offset2: float, slope2: float) -> Union[float, np.ndarray]:
```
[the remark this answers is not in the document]
peendebak: All 6 parameters need to be individual arguments, otherwise the `lmfit` package will not work. The `x` is the index of the experiment, I will rename.
The type of `x` is such that you can apply the method both for a range of indices (e.g. an array) or a scalar (e.g. a float).

src/tests/unittests/algorithms/test_allxy.py
```diff
@@ -0,0 +1,40 @@
...
+        self.assertIsInstance(result, dict)
+        np.testing.assert_array_almost_equal([0.1, 0, .5, 0, .9, 0], result['fitted_parameters'], decimal=1)
+
+        plot_allxy(dataset, result, fig=1, plot_initial_estimate=True)
+        plt.close(1)
+
+    def test_allxy_model(self):
+        for offset in [0, .1, .2]:
+            for slope in [-.1, 0, .1]:
+                for idx in [0, 1, 2, 3, 4]:
```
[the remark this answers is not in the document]
peendebak: Not sure how that would look like...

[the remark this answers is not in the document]
CumulonimbusCalvus: Yes, this was indeed not very trivial. This worked for me:
```
import itertools

for offset, slope, idx in itertools.product([.0, .1, .2], [-.1, 0, .1], [0, 1, 2, 3, 4]):
   ...
```

src/qtt/algorithms/allxy.py
```diff
@@ -0,0 +1,125 @@
...
+    ground_state_rotations = ['II', 'XX', 'YY', 'XY', 'YX']
+    equator_state_rotations = ['xI', 'yI', 'xy', 'yx', 'xY', 'yX', 'Xy', 'Yx', 'xX', 'Xx', 'yY', 'Yy']
+    excited_state_rotations = ['XI', 'YI', 'xx', 'yy']
+    allxy_combinations_input = ground_state_rotations + equator_state_rotations + excited_state_rotations
+    allxy_combinations = [(xymapping[x[0]], xymapping[x[1]]) for x in allxy_combinations_input]
+
+    return allxy_combinations
+
+
+def allxy_model(indices: Union[float, np.ndarray], offset0: float, slope0: float, offset1: float, slope1: float, offset2: float, slope2: float) -> Union[float, np.ndarray]:
```
CumulonimbusCalvus: You don't want to convert the arguments `slope0`, `slope1` and `slope2` into a single tuple argument `slopes`? Same for `offset0`, `offset1` and `offset2` into 'offsets'?

peendebak: Grouping parameters does not does work for the `lmfit` module. We could do a 
```
def _allxy_model(index, offset0, offset1, offset2, slope0, slope1, slope2):
   ....
   ...
   return calculated_value

def allxy_model(index, offsets, slopes):
   return _allxy_model(index, *offsets, *slopes)
```
In that way the public interface is `def allxy_model(index, offsets, slopes)`, but internally we can calculate with `lmfit` using `_allxy_model`. Alternatively we could do
```
def _allxy_model(index, offset0, offset1, offset2, slope0, slope1, slope2):
   ....
   ...
   return calculated_value

def allxy_model(index, parameters):
   return _allxy_model(index, *parameters)
```
In qtt we do both, e.g. compare https://github.com/QuTech-Delft/qtt/blob/dev/src/qtt/algorithms/functions.py#L13 or https://github.com/QuTech-Delft/qtt/blob/dev/src/qtt/algorithms/functions.py#L80

@CumulonimbusCalvus @QFer I don't really mind which convention to use, we can convert it either way. For this particular case I would pick either
```
def allxy_model(index, parameters): 
```
or
```
def _allxy_model(index, offset0, offset1, offset2, slope0, slope1, slope2)
```

CumulonimbusCalvus: Ah, now I see why this is needed. The `parameters` option makes most sense to me, as long if the `parameters` argument is well documented.

peendebak:
@CumulonimbusCalvus You did not refactor the code to have the interface as `def allxy_model(index, parameters)`. Will you pick this up in another PR or leave it as is?

Changing it is not backwards compatible, so should be done before the next release.

PR #665, opened 2019-10-04
[OP-85] fix negative frequency in gauss_ramsey fit

mjwoudstra:


mjwoudstra:
The build fails with a seemingly unrelated error.
@CumulonimbusCalvus , @peendebak : any idea what this is and how to fix it?

[NbConvertApp] ERROR | Error while converting 'docs/notebooks/analysis/example_charge_sensor.ipynb'

----> 1 h = 1e9*scipy.constants.h/(1e-6*scipy.constants.elementary_charge)  # Planck's constant in units [ueV/GHz]

      2 kb = scipy.constants.k/(1e-6*scipy.constants.elementary_charge)  # [ueV/K], Boltzmann constant

      3 kT = 10e-3 * kb  # effective electron temperature in ueV

AttributeError: module 'scipy' has no attribute 'constants'

peendebak:
> The build fails with a seemingly unrelated error.
> @CumulonimbusCalvus , @peendebak : any idea what this is and how to fix it?
> 
> [NbConvertApp] ERROR | Error while converting 'docs/notebooks/analysis/example_charge_sensor.ipynb'
> 
> ----> 1 h = 1e9_scipy.constants.h/(1e-6_scipy.constants.elementary_charge) # Planck's constant in units [ueV/GHz]
> 
> ```
>   2 kb = scipy.constants.k/(1e-6*scipy.constants.elementary_charge)  # [ueV/K], Boltzmann constant
> 
>   3 kT = 10e-3 * kb  # effective electron temperature in ueV
> ```
> 
> AttributeError: module 'scipy' has no attribute 'constants'

You removed `import scipy.constants` from one of the files. This triggered an import error in one of the notebooks. I fixed the import in the notebook. @mjwoudstra

src/qtt/algorithms/functions.py
```diff
@@ -340,24 +341,46 @@ def cost_function(params):
...
+    """
+    A sine wave with exponential decay of a power of x.
+    Function:
+    $$ amplitude * exp(-(x_data*decay_rate)**power) * sin(2pi*frequency * x_data + phase) + offset $$
+    Function parameters:
+
+    Args:
+        x_data (np.ndarray): array of independent variable values where the function needs to be evaluated.
+        amplitude: amplitude of the sine at x=0.
+        decay_rate (Hz): value of 1/x where the damping factor is 1/e
```
peendebak: The type of `decay_rate` is not Hz.
```suggestion
        decay_rate (float): value of 1/x where the damping factor is 1/e. Unit is Hz
```

src/qtt/algorithms/functions.py
```diff
@@ -340,24 +341,46 @@ def cost_function(params):
...
+        power (-): power of x_data*decay_rate inside the exponent.
+        frequency (Hz)
```
peendebak:
```suggestion
        frequency (float): Frequency in Hz
```
Similar remarks for the other parameters

src/qtt/algorithms/functions.py
```diff
@@ -340,24 +341,46 @@ def cost_function(params):
...
+        phase (radians): phase of the sine at x=0.
+        offset: constant offset of the function.
+
+    Returns:
+        np.ndarray with values of function for all x_data.
```
peendebak:
```suggestion
        Array with values of function for all x_data.
```

src/qtt/algorithms/functions.py
```diff
@@ -340,24 +341,46 @@ def cost_function(params):
...
     by a sinus.
     Function as used by T.F. Watson et all., example in qtt/docs/notebooks/example_fit_ramsey.ipynb
 
-    $$ gauss_ramsey = A * exp(-(x_data/t2s)**2) * sin(2pi*ramseyfreq * x_data - angle) +B  $$
+    $$ gauss_ramsey = A * exp(-(x_data/t2s)**2) * sin(2pi*ramsey_freq * x_data - angle) +B  $$
 
     Args:
-        x_data (array): the data for the input variable
-        params (array): parameters of the gauss_ramsey function, [A,t2s,ramseyfreq,angle,B]
+        x_data (numpy array): the data for the input variable
```
peendebak: Why write `numpy array` here and `array` in the output? Be consistent if possible.

src/qtt/algorithms/functions.py
```diff
@@ -480,26 +513,60 @@ def func(params): return cost_gauss_ramsey(x_data, y_data, params, weight_power=
...
+        x_data: Input array with time variable (seconds)
         y_data: Input array with signal
         fit_parameters: Result of fit_gauss_ramsey
+        fig: figure number to plot into
+        initial_parameters: if given, also plot function with initial parameters
+        show_parameters: if True, then all parameter values are shown in the plot
     """
-    test_x = np.linspace(0, np.max(x_data), 200)
-    freq_fit = abs(fit_parameters[2] * 1e-6)
+    def get_fit_pars_string(fit_parameters: List[float] = fit_parameters, title: str = "Fit parameters:") -> str:
```
peendebak: We have the convention in qtt to in a single module to either i) annotate all methods and classes with type annotations or ii) don't do any type annotation at all.

We might want to re-evaluate that decision. @QFer @mjwoudstra 
In any case, we need to decide how to annotate things like fit parameters. It would be something like `Union[ List[float], np.ndarray]`, but that is not so pretty.

src/qtt/algorithms/functions.py
```diff
@@ -340,24 +341,46 @@ def cost_function(params):
     return fitted_parameters
 
 
+def damped_sine_wave(x_data, amplitude, decay_rate, power, frequency, phase, offset):
+    """
+    A sine wave with exponential decay of a power of x.
+    Function:
+    $$ amplitude * exp(-(x_data*decay_rate)**power) * sin(2pi*frequency * x_data + phase) + offset $$
+    Function parameters:
```
peendebak:
```suggestion

```

src/qtt/algorithms/functions.py
```diff
@@ -381,33 +404,43 @@ def estimate_dominant_frequency(signal, sample_rate=1, remove_dc=True, fig=None)
...
+            w[0] = 0
 
     ff = freqs[np.argmax(np.abs(w))]
+    ff_pos = freqs_pos[np.argmax(np.abs(w_pos))]
 
     if fig:
         plt.figure(fig)
         plt.clf()
-        plt.plot(freqs, np.abs(w), '.b')
+        plt.plot(freqs, np.abs(w), '.r')
```
peendebak:
```suggestion
        plt.plot(freqs, np.abs(w), '.r', label='FFT')
```

src/qtt/algorithms/functions.py
```diff
@@ -381,33 +404,43 @@ def estimate_dominant_frequency(signal, sample_rate=1, remove_dc=True, fig=None)
...
+        plt.plot(freqs_pos, np.abs(w_pos), '.b')
         plt.xlabel('Frequency')
         plt.ylabel('Abs of fft')
-        plot_vertical_line(ff)
-    return ff
+        plot_vertical_line(ff, color='r')
```
peendebak:
```suggestion

```
Not not plot this value. It is not used.

src/qtt/algorithms/functions.py
```diff
@@ -381,33 +404,43 @@ def estimate_dominant_frequency(signal, sample_rate=1, remove_dc=True, fig=None)
...
+        plot_vertical_line(ff_pos, color='b')
```
peendebak:
```suggestion
        plot_vertical_line(ff_pos, color='b', label='Estimated dominant frequency')
```

src/qtt/algorithms/functions.py
QFer: Typehinting a module as a whole is my preference because it's much more efficient if one person typehints a module instead of more persons doing it bit by bit, especially the more difficult types.
Further, it can then be checked by mypy as a whole.

src/qtt/algorithms/functions.py
mjwoudstra: good catch

src/qtt/algorithms/functions.py
mjwoudstra: for the PyCharm type checker, `array` is different from `numpy array` (or `np.ndarray`), so I tried to put the proper type. We can use `np.ndarray` everywhere instead of `numpy array` (indeed good to be consistent there)

src/qtt/algorithms/functions.py
mjwoudstra: I did not want type type annotate the whole module, so let's just remove them from this one function.

mjwoudstra: I did not want type type annotate the whole module, so let's just remove them from this one function.

src/qtt/algorithms/functions.py
mjwoudstra: Then we do not need to calculate it either (few lines above)

peendebak:
Closed. Functionality has been implemented partly elsewhere.

PR #666, opened 2019-10-14
Restructure fitFermiLinear

peendebak:
* Fix plotting of fitFermiLinear (use lever arm)
* Create plot method for fitFermiLinear
* output lever_arm in results
* Rename argument `l` to `lever_arm` in `fitFermiLinear`

### DIFF
--- a/src/qtt/algorithms/fitting.py
+++ b/src/qtt/algorithms/fitting.py
@@ -154,7 +154,7 @@ def initFermiLinear(x_data, y_data, fig=None):
 
 # %%
 
-def fitFermiLinear(x_data, y_data, verbose=0, fig=None, l=1.16, use_lmfit=False):
+def fitFermiLinear(x_data, y_data, verbose=0, fig=None, lever_arm = 1.16, l=None, use_lmfit=False):
     """ Fit data to a Fermi-Linear function
 
     Args:
@@ -162,7 +162,7 @@ def fitFermiLinear(x_data, y_data, verbose=0, fig=None, l=1.16, use_lmfit=False)
         y_data (array): dependent variable data
         verbose (int) : verbosity (0 == silent). Not used
         fig (int) : figure number
-        l (float): leverarm passed to FermiLinear function
+        lever_arm (float): leverarm passed to FermiLinear function
         use_lmfit (bool): If True use lmfit for optimization, otherwise use scipy
 
     Returns:
@@ -174,6 +174,10 @@ def fitFermiLinear(x_data, y_data, verbose=0, fig=None, l=1.16, use_lmfit=False)
     xdata = np.array(x_data)
     ydata = np.array(y_data)
 
+    if l is not None:
+        warnings.warn('use argument lever_arm instead of l')
+        lever_arm = l
+        
     # initial values
     linear_part, fermi_part = initFermiLinear(xdata, ydata, fig=None)
     initial_parameters = linear_part + fermi_part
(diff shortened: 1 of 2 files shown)


src/qtt/algorithms/fitting.py
```diff
@@ -197,21 +201,38 @@ def fermi_linear_fitting_function(xdata, a, b, cc, A, T):
...
     if fig is not None:
-        y = FermiLinear(xdata, *list(fitted_parameters))
-        plt.figure(fig)
-        plt.clf()
-        plt.plot(xdata, ydata, '.b', label='data')
-        plt.plot(xdata, y, 'm-', label='fitted FermiLinear')
-        plt.legend(numpoints=1)
+        plot_FermiLinear(xdata, ydata, results, fig=fig)
+
+    return fitted_parameters, results
```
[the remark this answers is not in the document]
peendebak: Yes, backwards compatibility. Preferred style for any new methods is to return only a dictionary.

src/qtt/algorithms/fitting.py
```diff
@@ -154,15 +154,15 @@ def initFermiLinear(x_data, y_data, fig=None):
...
-def fitFermiLinear(x_data, y_data, verbose=0, fig=None, l=1.16, use_lmfit=False):
+def fitFermiLinear(x_data, y_data, verbose=0, fig=None, lever_arm=1.16, l=None, use_lmfit=False):
     """ Fit data to a Fermi-Linear function
 
     Args:
         x_data (array): independent variable data
         y_data (array): dependent variable data
         verbose (int) : verbosity (0 == silent). Not used
         fig (int) : figure number
-        l (float): leverarm passed to FermiLinear function
```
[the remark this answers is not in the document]
peendebak: I documented it as a deprecated parameter. It is not clear when this will be deleted. For entire functions we have the `rdeprecated` decorator, which can be used to give a message untill when the function is supported. For single arguments we do not have such a thing.

src/qtt/algorithms/fitting.py
```diff
@@ -171,20 +172,20 @@ def fitFermiLinear(x_data, y_data, verbose=0, fig=None, lever_arm=1.16, l=None,
...
 
     # initial values
-    linear_part, fermi_part = initFermiLinear(xdata, ydata, fig=None)
+    linear_part, fermi_part = initFermiLinear(x_data, y_data, fig=None)
     initial_parameters = linear_part + fermi_part
 
     # fit
-    def fermi_linear_fitting_function(xdata, a, b, cc, A, T):
-        return FermiLinear(xdata, a, b, cc, A, T, l=lever_arm)
+    def fermi_linear_fitting_function(x_data, a, b, cc, A, T):
```
QFer: The same comment is still valid. You are shadowing a variable in outer scope. x_data is already used in the outer method. You have to look twice to know to which x_data in line 188 refers to.
Because the only thing is that FermiLinear is called with an extra argument. Maybe you could use partial instead  partial(FermiLinear, l=lever_arm) instead of fermi_linear_fitting_function?

peendebak: Ah I see the point now. I will use `partial`

peendebak: The `partial` does not work, because `partial(FermiLinear, l=lever_arm)` does not have `__name__` which is required by lmfit.

I am therefore still using the old imlementation, which seems the cleanest solution.

PR #667, opened 2019-10-17
[OP-42] Preparation for release qtt

QFer:
Setup procedure changed:
* Changed version to 1.1.3 for new release
* Only necessary packages are installed from setup. Both for qtt and for generating readthedocs
* Extra_require in setup for documentation (instead of requirements file)
* Polygon3 binaries for python 3.7 temporary in repo
* requirements_lock.txt for installing tested dependencies
* No requirements file for installing qtt
* README.md changed for the new installation
* travis automatic update to pypi from master
* Changelog updated


peendebak:
@QFer Since `condalist.yml` is removed from qtt, also the instructions at `https://github.com/VandersypenQutech/spin-projects/blob/master/INSTALL.md` need to be updated.

README.md
```diff
@@ -17,24 +17,76 @@ on [Qcodes](https://github.com/qdev-dk/Qcodes) (basic framework such as instrume
...
 ```
+$ git clone https://github.com/QuTech-Delft/qtt.git
+$ cd qtt
 $ pip install -e .
 ```
+### When problems arise
+Sometimes the default installation does not work because of incompatible dependencies between the used packages
+on your system. To be sure you use all the right versions of the packages used by QTT and its dependencies do:
+```
+$ pip install -r requirements_lock.txt
```
CumulonimbusCalvus:
```suggestion
$ pip install . -r requirements_lock.txt
```

setup.py
```diff
@@ -23,25 +24,36 @@ def get_version(verbose=1, filename='src/qtt/version.py'):
...
+
+if platform.python_version() < "3.7.0":
+    install_requires.append('Polygon3')
+else:
+    if platform.system() == 'Windows':
+        # When Polygon3 not yet in git-repository get it locally
+        file_path = abspath(getsourcefile(lambda: 0))
+        bin_dir = os.path.join(os.path.dirname(file_path), 'bin')
+        Polygon3_local = f'Polygon3 @ file://{bin_dir}/Polygon3-3.0.8-cp37-cp37m-win_amd64.whl'
+        Polygon3_git = 'Polygon3 @ https://github.com/QuTech-Delft/qtt/bin/Polygon3-3.0.8-cp37-cp37m-win_amd64.whl'
```
CumulonimbusCalvus: This does not work for me on both Windows. The local wheel does work for me. Should we just use that for now?

README.md
```diff
@@ -17,24 +17,76 @@ on [Qcodes](https://github.com/qdev-dk/Qcodes) (basic framework such as instrume
...
+### Setting up a virtual environment
+To create a clean virtual Python environment for your qtt development do:
+```
+$ mkdir qtt
+$ cd qtt
+$ python -m venv env
+```
+Now activate the virtual environment. On Linux do:
+```
+$ . ./env/Scripts/activate
```
CumulonimbusCalvus: I also tried it on ubuntu with a virtual machine. For it to work I needed to do the following:
```
sudo apt install git python3 python3-venv python3-dev gcc
python -m venv env
git clone git@github.com:qutech-delft.qtt.git
pip install wheel
pip install -e .
```

Should we add this as well?

README.md
QFer: Added the dot

README.md
QFer: I added a sentence that software as python and git should be installed already (with a link to download pages). I added the pip wheel install to the description

CumulonimbusCalvus:
> @QFer Since condalist.yml is removed from qtt, also the instructions at https://github.com/VandersypenQutech/spin-projects/blob/master/INSTALL.md need to be updated.

Will be updated in OP-103.

PR #673, opened 2019-10-23
[OP-56 prepare] update polyintersect 

peendebak:
* Update docstring
* Add tests for polyintersect


src/tests/unittests/test_pgeometry.py
```diff
@@ -62,12 +62,33 @@ def test_directionMean(self):
...
+        self.assertEmptyPolygon(intersection_polygon)
+
+    def test_polyintersect_identical_polygons(self):
+        x1 = np.array([(0, 0), (1, 1), (1, 0)])
+        intersection_polygon = pgeometry.polyintersect(x1, x1)
+        for pt in x1:
+            self.assertIn(pt, intersection_polygon)
+
+    def test_polyintersect_degenerate_polygons(self):
+        x1 = np.array([(0, 0)])
```
QFer: Raises an exception in Polygon3: cPolygon.Error: Invalid polygon or contour for operation
Should be catched, now the unit test fails.

peendebak:
> I thought you were busy with a replacement for Polygon3. It's not in this PR?

The replacement is not fast enough, we'll have to look at other options.

src/tests/unittests/test_pgeometry.py
peendebak: @QFer Test is fixed. It was a little bit tricky because the exception raises was not a standard exception.

src/tests/unittests/test_pgeometry.py
```diff
@@ -62,12 +63,36 @@ def test_directionMean(self):
...
+        for pt in x1:
+            self.assertIn(pt, intersection_polygon)
+
+    @staticmethod
+    def test_polyintersect_degenerate_polygons():
+        x1 = np.array([(0, 0)])
+        x2 = np.array([(1, 0), (1.5, 1.5)])
+        try:
+            pgeometry.polyintersect(x1, x2)
+        except PolygonError: # the assertRaises does not work due to the ctypes error message
```
QFer: Nothing is tested in this test. It never fails it never succeeds. What is the goal of this test?

peendebak: I improved the test. A degenerate polygon should either lead to an empty intersection or raise an error. @QFer

PR #678, opened 2019-10-31
[OP-102] Angle estimation

peendebak:
* Fix angle estimation in `estimate_parameters_damped_sine_wave`
* Fix bug in `sample_rate estimation` of `estimate_parameters_damped_sine_wave`
* Use subpixel estimation in dominant frequency estimate


src/qtt/algorithms/functions.py
```diff
@@ -473,14 +480,17 @@ def estimate_parameters_damped_sine_wave(x_data, y_data, exponent=2):
...
     if A == 0:
         angle = 0
     else:
         n_start = max(min((y_data[0] - B) / A, 1), -1)
-        angle = -np.arcsin(n_start)
-    t2s = 2 * duration / decay_factor
+        angle_first_datapoint = -np.arcsin(n_start)
+        angle = angle_first_datapoint + 2 * np.pi * ramseyfreq * x_data[0]
+        angle = np.mod(np.pi + angle, 2 * np.pi) - np.pi
+    t2s = 2 * duration / (decay_factor)
```
QFer: the braces around decay factor are not needed

src/qtt/algorithms/functions.py
```diff
@@ -473,14 +480,17 @@ def estimate_parameters_damped_sine_wave(x_data, y_data, exponent=2):
     decay_factor = (mean_left + laplace_factor) / (mean_right + laplace_factor)
 
     duration = x_data[-1] - x_data[0]
-    sample_rate = x_data.size / duration
+    sample_rate = (x_data.size - 1) / duration
     ramseyfreq = estimate_dominant_frequency(y_data, sample_rate=sample_rate)
```
QFer: maybe rename to ramsey_freq

src/qtt/algorithms/functions.py
QFer: I would like to review the math in your pull request if you want but the reference in the function heading doesn't give me enough information to do that. Maybe describing in docstring more what happens may make this easier (I like the extensive descriptions in numpy that really help deciding if the function is accurate to use for the job or not)

src/qtt/algorithms/generic.py
```diff
@@ -92,8 +95,13 @@ def subpixelmax(A, mpos, verbose=0):
 
     cy = val
     ay = (valm + valp) / 2 - cy
-    by = ay + cy - valm
-    shift = -by / (2 * ay)  # Maxima of quadradic
+
```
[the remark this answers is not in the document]
peendebak: Yes, it is ok. We have to make a choice on how to treat these corner cases. Other options would be to raise an error, or just select 0 (or `dsize-1`). That last option might even be better, but is out of scope of this PR.

src/qtt/algorithms/functions.py
```diff
@@ -435,17 +436,23 @@ def estimate_dominant_frequency(signal, sample_rate=1, remove_dc=True, fig=None)
 
     if remove_dc:
         w[0] = 0
+    w[freqs < 0] = 0
 
-    ff = freqs[np.argmax(np.abs(w))]
+    dominant_idx = np.argmax(np.abs(w))
+    dominant_frequency = freqs[dominant_idx]
+
+    if dominant_idx > 0 and dominant_idx < freqs.size / 2 - 1:
```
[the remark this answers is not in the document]
peendebak: Yes, and that is a bit more elegant so I will use it. Btw: you can use github suggestions for these changes, I can accept those right away.

src/qtt/algorithms/generic.py
```diff
@@ -92,8 +95,13 @@ def subpixelmax(A, mpos, verbose=0):
...
+    if np.any(ay==0):
+        # degenerate case
+        raise Exception('position specified by mpos is not a maximum')
```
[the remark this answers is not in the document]
peendebak: In case of 3 equal values we have a maximum, but not a strict maximum. The calculation would fail (division by `ay`), so that is why I added the exception. With a little more thought I think that is more elegant so solve this by then just leaving the maximum as is, instead of raising the error.

src/qtt/algorithms/functions.py
```diff
@@ -435,17 +436,23 @@ def estimate_dominant_frequency(signal, sample_rate=1, remove_dc=True, fig=None)
 
     if remove_dc:
         w[0] = 0
+    w[freqs < 0] = 0
```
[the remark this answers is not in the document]
peendebak: `w` is the fft of the input signal. I don't want to use `fft` (since that is already the method performing the transform). Would `signal_fft` be better?

src/qtt/algorithms/generic.py
```diff
@@ -93,13 +97,16 @@ def subpixelmax(A, mpos, verbose=0):
...
     by = ay + cy - valm
-    shift = -by / (2 * ay)  # Maxima of quadradic
+
+    if np.any(ay == 0):
+        shift = 0 * ay
+    else:
+        shift = -by / (2 * ay)  # Maxima of quadradic
 
     if verbose:
         print('subpixelmax: mp %d, pp %d\n', mp, pp)
```
QFer: The print statement is not correct:
```suggestion
        print('subpixelmax: mp %d, pp %d\n' % (mp, pp))
```

src/qtt/algorithms/generic.py
```diff
@@ -93,13 +97,16 @@ def subpixelmax(A, mpos, verbose=0):
...
         print('subpixelmax: ap %.3f, by %.3f , cy %.3f\n', ay, by, cy)
```
QFer:
```suggestion
        print('subpixelmax: ap %.3f, by %.3f , cy %.3f\n' % (ay, by, cy))
```

QFer: There is another print. I cannot make a suggestion for it but it should be: 

print('subpixelmax1d: shift %.3f\n' % shift)

src/qtt/algorithms/functions.py
```diff
@@ -473,16 +490,19 @@ def estimate_parameters_damped_sine_wave(x_data, y_data, exponent=2):
     decay_factor = (mean_left + laplace_factor) / (mean_right + laplace_factor)
 
     duration = x_data[-1] - x_data[0]
-    sample_rate = x_data.size / duration
-    ramseyfreq = estimate_dominant_frequency(y_data, sample_rate=sample_rate)
+    sample_rate = (x_data.size - 1) / duration
+    frequency = estimate_dominant_frequency(y_data, sample_rate=sample_rate)
+
     if A == 0:
```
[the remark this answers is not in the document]
peendebak: In this case it doesn't really matter. If A is close to zero than the other branch is just as good (or better) as the A==0 branch (we just cannot divide by zero as that raises an error)

PR #679, opened 2019-11-09
add example notebook for mongodb

peendebak:
* Add example notebook for StorageInterface functionality


QFer:
Pieter, when do you think this notebook is done and can be reviewed? Currently there is some parameter mismatch with qilib.

peendebak:
> Pieter, when do you think this notebook is done and can be reviewed? Currently there is some parameter mismatch with qilib.

@QFer When there has been a new release of qilib on pypi.

PR #680, opened 2019-11-11
[OP-96] add option to refit double gaussian

peendebak:
* Improve fitting a double gaussian by initialisation of a previous fit


src/qtt/algorithms/functions.py
```diff
@@ -215,13 +215,49 @@ def _double_gaussian(x, A_dn, A_up, sigma_dn, sigma_up, mean_dn, mean_up):
...
+
+    Args:
+        result_dict(dict): Result dictionary from fit_double_gaussian
+        x_data (array): Independent data
+        y_data (array): Signal data
+        gaussian_amplitude_ratio_threshold (float): If ratio between amplitudes of Gaussian peaks is larger than this fit, re-estimate
+    Returns:
+        Dictionary with improved fitting results
+    """
+    if result_dict['left'][2] > result_dict['right'][2]:
```
QFer: Instead of all the indexes [0], [1], [2] which actually stands for mean, std and amplitude you can do something like:
```
mean = operator.itemgetter(0)
std = operator.itemgetter(1)
amplitude = operator.itemgetter(2)
```
and then you may do:
```
if amplitude(result_dict['left']) > amplitude(result_dict['right']):
..
gaussian_ratio = amplitude(large_gaussian_parameters) / amplitude(small_gaussian_parameters)
..
        idx = np.logical_and(x_data > mean(large_gaussian_parameters) - 1.5 * std(large_gaussian_parameters),
                             x_data < mean(large_gaussian_parameters) + 1.5 * std(large_gaussian_parameters))
```
which to me is more clear than the hard indexing.

src/qtt/algorithms/functions.py
```diff
@@ -215,13 +215,49 @@ def _double_gaussian(x, A_dn, A_up, sigma_dn, sigma_up, mean_dn, mean_up):
...
+
     result_dict['left'] = np.take(par_fit, [4, 2, 0])
     result_dict['right'] = np.take(par_fit, [5, 3, 1])
     result_dict['type'] = 'fitted double gaussian'
 
     return par_fit, result_dict
 
 
+def refit_double_gaussian(result_dict, x_data, y_data, gaussian_amplitude_ratio_threshold=8):
+    """ Improve fit of double Gaussian by estimating the initial parameters based on an existing fit
```
QFer: I know broadly what is happening in this method but I find it hard to conclude that all the mathematics is correct.

src/qtt/algorithms/functions.py
```diff
@@ -215,13 +215,49 @@ def _double_gaussian(x, A_dn, A_up, sigma_dn, sigma_up, mean_dn, mean_up):
...
+
+    Args:
+        result_dict(dict): Result dictionary from fit_double_gaussian
```
QFer: strange name for an input parameter, the _dict part is superfluous I guess.

src/qtt/algorithms/functions.py
```diff
@@ -215,13 +215,49 @@ def _double_gaussian(x, A_dn, A_up, sigma_dn, sigma_up, mean_dn, mean_up):
...
+    result_dict['reduced_chi_squared'] = result.redchi
+
     result_dict['left'] = np.take(par_fit, [4, 2, 0])
     result_dict['right'] = np.take(par_fit, [5, 3, 1])
     result_dict['type'] = 'fitted double gaussian'
 
     return par_fit, result_dict
 
 
+def refit_double_gaussian(result_dict, x_data, y_data, gaussian_amplitude_ratio_threshold=8):
```
QFer: Can this method be called without first calling fit_double_gaussian? If not, why not call fit_double_gaussian from within this method (the refit... part of the name is not accurate then of course).

src/qtt/algorithms/functions.py
peendebak: That is really cool!

src/qtt/algorithms/functions.py
peendebak: I agree. Changing it is (a little bit) backwards incompatible

src/qtt/algorithms/functions.py
peendebak: I cannot be called without first calling `fit_double_gaussian` (unless someone prepares data similar to the output of that method). However, if build this into the method, then a user has less freedom in calling the method. The current approach seems the most modular to me

src/qtt/algorithms/functions.py
peendebak: Any suggestions on how to improve? I split part of the function and the other suggestions also helped to clear up the code.

PR #684, opened 2019-11-19
add metadata if no station present

peendebak:
When adding notes to a PPT slide one can pass a `DataSet`. Fields such as `station`, `scanjob` and others are then converted to notes. This PR makes sure notes are created, even when the `station` field is not present.


peendebak:
@QFer Addresssed the comments by extending the tests a bit.

PR #687, opened 2019-11-21
remove code that points to spin-qubits repo

peendebak:
The code is not public, and has not updated for quite some time.


peendebak:
> I don't have comments on the change itself, I think its oke but now in the block preceding the deleted block (# %% Make a 2D scan) we prepare a scanjob and are doing a 2Dscan for nothing. The resulting data isn't used. I think we can also delete that portion of the code when there are no (unwanted) side effects calling this 2Dscan.

@QFer The data is indeed not used, but the code should be there as it is an example of how to perform a 2D scan.

PR #692, opened 2019-11-29
[OP-103] Fixes for transition of m4i to qcodes contrib

peendebak:
Since the M4i driver moved to qcodes contrib, we have to update some checks in the code.


src/qtt/measurements/scans.py
```diff
@@ -1948,6 +1946,21 @@ def acquire_segments(station, parameters, average=True, mV_range=2000,
 
     return alldata
 
+def _is_m4i(instrument):
+    try:
+        is_m4i = isinstance(instrument, qcodes.instrument_drivers.Spectrum.M4i.M4i)
+        if is_m4i:
+               warnings.warn('please use qcodes_contrib_drivers.drivers.Spectrum.M4i instead of qcodes.instrument_drivers.Spectrum.M4i.M4i)')
```
QFer: line too long
```suggestion
               warnings.warn(f'please use qcodes_contrib_drivers.drivers.Spectrum.M4i'
                             f' instead of qcodes.instrument_drivers.Spectrum.M4i.M4i')
```

src/qtt/measurements/scans.py
```diff
@@ -1948,6 +1946,21 @@ def acquire_segments(station, parameters, average=True, mV_range=2000,
 
     return alldata
 
+def _is_m4i(instrument):
```
QFer: two newlines between methods

src/qtt/measurements/scans.py
```diff
@@ -1948,6 +1946,21 @@ def acquire_segments(station, parameters, average=True, mV_range=2000,
...
+    try:
```
QFer: why the try clause. Isinstance is just returning True or False, only when qcodes.instrument_drivers.Spectrum.M4i.M4i is not a class you can expect a TypeError

src/qtt/measurements/scans.py
```diff
@@ -1948,6 +1946,21 @@ def acquire_segments(station, parameters, average=True, mV_range=2000,
...
+        is_m4i = isinstance(instrument, qcodes.instrument_drivers.Spectrum.M4i.M4i)
+        if is_m4i:
+               warnings.warn('please use qcodes_contrib_drivers.drivers.Spectrum.M4i instead of qcodes.instrument_drivers.Spectrum.M4i.M4i)')
+               return is_m4i
+    except:
```
QFer: you should not use a bare except: when not needed
Depending on an earlier comment this should be changed to catch only TypeError

src/qtt/measurements/scans.py
```diff
@@ -1948,6 +1946,21 @@ def acquire_segments(station, parameters, average=True, mV_range=2000,
...
+        is_m4i = False
+    try:
+        import qcodes_contrib_drivers.drivers.Spectrum.M4i
+        is_m4i = isinstance(instrument, qcodes_contrib_drivers.drivers.Spectrum.M4i.M4i)
+    except:
```
QFer: also here again a bare Except, catch ImportError, TypeError when needed.

src/qtt/measurements/scans.py
```diff
@@ -1948,6 +1946,21 @@ def acquire_segments(station, parameters, average=True, mV_range=2000,
...
+        is_m4i = False
+
+    return is_m4i
```
QFer: Leave two newlines until next method

src/qtt/measurements/scans.py
```diff
@@ -1811,10 +1811,8 @@ def measuresegment(waveform, Naverage, minstrhandle, read_ch, mV_range=2000, pro
     if device_parameters is None:
         device_parameters = {}
         
-    try:
-        is_m4i = isinstance(minstrhandle, qcodes.instrument_drivers.Spectrum.M4i.M4i)
-    except:
-        is_m4i = False
+    is_m4i = _is_m4i(minstrhandle)
+
     try:
```
QFer: if is_m4i is True it is not an uhfli or a scope reader. Why trying? The same for trying scope reader when minsthandle is a uhfli? A little bit of boyscouting needed here.

QFer: Often these if elif elif else constructions depending on the type of instance can be refactored to a proper design (inheritance structure with an overridden method measure_segment). What is your opinion on these type of changes.

CumulonimbusCalvus:
As mentioned I will update the PR, since @peendebak is on holiday. 
We would like to have this update in the new release.

src/qtt/measurements/scans.py
```diff
@@ -1946,21 +1943,28 @@ def acquire_segments(station, parameters, average=True, mV_range=2000,
...
-               return is_m4i
-    except:
-        is_m4i = False
+
+def _is_measurement_device(instrument_handle: Any, class_type: Type) -> bool:
+    """ Returns True if the instrument handle is of the given type, else False.
+
+        This function checks whether the given handle is of the correct instrument type.
+        All error's are catched related to importing of not installed drivers or instruments
+        which are note connected.
```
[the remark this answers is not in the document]
CumulonimbusCalvus: Changed it.

PR #694, opened 2019-12-16
Cleanup result selection of sensingdot_t

peendebak:


src/qtt/structures.py
```diff
@@ -347,6 +347,15 @@ def scan2D(sd, ds=90, stepsize=4, fig=None, verbose=1):
...
         return alldata
 
+    def _select_results(self, goodpeaks):
+        if len(goodpeaks) > 0:
+            self.sdval[1] = float(goodpeaks[0]['xhalfl'])
+            self.targetvalue = float(goodpeaks[0]['yhalfl'])
+            self._selected_peak = goodpeaks[0]
+        else:
+            self._selected_peak = None
+            raise qtt.exceptions.CalibrationException('fastTune: could not find good peak')
```
QFer: Because the function is also called from autoTune, the fastTune reference in the exception text can mislead the user.

PR #696, opened 2019-12-17
fix for period argument of sensingdot_t

peendebak:
* Pass `period` argument to measurement method
* Argument value set to 1-e3 for backwards compatibility


src/qtt/structures.py
```diff
@@ -464,6 +464,7 @@ def fastTune(self, Naverage=90, sweeprange=79, period=.5e-3, location=None,
                 scanjob['sweepdata'] = {'param': gate, 'start': cc -
                                         sweeprange / 2, 'end': cc + sweeprange / 2, 'step': 4}
 
+            scanjob['sweepdata']['period']=period
```
QFer: spaces around =

src/qtt/structures.py
```diff
@@ -428,7 +428,7 @@ def _measurement_post_processing(self, dataset):
 
         return dataset
 
-    def fastTune(self, Naverage=90, sweeprange=79, period=.5e-3, location=None,
+    def fastTune(self, Naverage=90, sweeprange=79, period=1e-3, location=None,
```
QFer: Please add period in the parameter docstring. Why is it used in this function?

PR #704, opened 2020-01-02
derive calibration exception from normal exception

peendebak:
According to the docs https://docs.python.org/3/library/exceptions.html any user defined exception should be based on the normal `Exception`


peendebak:
> The test in test_exceptions.py should also change in:
> self.assertTrue(issubclass(qtt.exceptions.CalibrationException, Exception))
> Of course the test currently succeeds but it should test against Exception.

Done!

PR #711, opened 2020-02-18
disable all output channels when restarting the videomode

peendebak:
When the AWG is still running when the VideoMode tool is started, then not all active channels are disabled. This causus problems if the channels involved in the VideoMode do not cover all channels running on the AWG. This PR fixes that by disabling all output channels.

@sldesnoo-Delft


src/qtt/instrument_drivers/virtualAwg/virtual_awg.py
```diff
@@ -182,15 +182,19 @@ def disable_outputs(self, gate_names):
             A start command is required to enable the outputs.
 
         Arguments:
-            gate_names (list) The names of the gates which needs to be disabled.
+            gate_names (list or None) The names of the gates which needs to be disabled. If None, then disable all gates
```
QFer: Make argument gate_names optional and defaults to gate_names=None, then we don't have to call with parameter None.
Then it is also consistent with the disable_output implementations of the instruments itself

src/qtt/instrument_drivers/virtualAwg/virtual_awg.py
```diff
@@ -182,15 +182,19 @@ def disable_outputs(self, gate_names):
...
-        if VirtualAwg.__digitizer_name in self._settings.awg_map:
-            gate_names.extend([VirtualAwg.__digitizer_name])
-        if VirtualAwg.__awg_slave_name in self._settings.awg_map:
-            gate_names.extend([VirtualAwg.__awg_slave_name])
-        for name in gate_names:
-            (awg_number, channel_number, *_) = self._settings.awg_map[name]
-            self.awgs[awg_number].disable_outputs([channel_number])
+        if gate_names is None:
+            for awg in self.awgs:
+                awg.disable_outputs(None)
```
QFer: None not needed here anymore when the other comment is resolved.
```suggestion
                awg.disable_outputs()
```

src/qtt/measurements/videomode_processor.py
```diff
@@ -270,11 +270,7 @@ def stop(self):
                 self.station.RF.off()
             if hasattr(self.station, 'virtual_awg'):
                 self.station.virtual_awg.stop()
-                if isinstance(self.sweepparams[0], dict):
-                    keys = [list(item.keys())[0] for item in self.sweepparams]
-                    self.station.virtual_awg.disable_outputs(keys)
-                elif isinstance(self.sweepparams, str):
-                    self.station.virtual_awg.disable_outputs([self.sweepparams])
+                self.station.virtual_awg.disable_outputs(None)
```
QFer: We can call without None
```suggestion
                self.station.virtual_awg.disable_outputs()
```

PR #720, opened 2020-02-28
Fix the gain methods of the ZI HDAWG8

peendebak:
The range of the HDAWG8 device specifies the final output range of the device channels. For example a range of 3.0 means that the device has output range of -1.5 to +1.5 V. This corresponds to a gain of 1.5. In general `gain = range /2`

This PR fixes a bug in the ZI HDAWG8 driver where the gain and range where confused.


YakBizzarro:
Just a comment on that, since I've seen it passing by. Actually the range coincide with the gain, but under the assumption that your load is high impedance. If the load is 50 Ohm, the voltage is divided by two, since also the HDAWG output impedance is 50 Ohm.

peendebak:
@YakBizzarro Thanks for the info. I will add this to the documentation for our code.

src/tests/unittests/instrument_drivers/virtualAwg/awgs/test_ZurichInstrumentsHDAWG8.py
```diff
@@ -71,14 +71,14 @@ def test_retrieve_sampling_rate(self):
...
+        calls = [call.set('sigouts_{}_range'.format(ch), 1.0) for ch in range(8)]
         self.awg.assert_has_calls(calls)
 
     def test_retrieve_gain(self):
         self.awg.get.return_value = 0.2
-        self.assertEqual(0.2, self.zi_hdawg8.retrieve_gain())
+        self.assertEqual(0.1, self.zi_hdawg8.retrieve_gain())
 
-        with self.assertRaises(ValueError):
+        with self.assertRaises(Exception):
```
QFer: Here we're trying to get the ValueError raised in retrieve_gain when not all gains are the same. The reason you got an other exception than expected was because you now divide by 2 the value you got back from the lambda function that is defined here. So you get a string by int division error. So you are not testing what was meant in the old test.
When you change the lambda function for the side effect to:
self.awg.get.side_effect = lambda v: int(v[8:9])
we get the exception we want
```suggestion
        with self.assertRaises(ValueError):
```

src/tests/unittests/instrument_drivers/virtualAwg/awgs/test_ZurichInstrumentsHDAWG8.py
```diff
@@ -71,14 +71,14 @@ def test_retrieve_sampling_rate(self):
...
             self.awg.get.side_effect = lambda v: v
```
QFer:
```suggestion
            self.awg.get.side_effect = lambda v: int(v[8:9])
```

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - fix for period argument of sensingdot_t (#696)
 - Fix create_vectorscan for the new virtual awg  (#702)
 - fix deprecation warnings (#703)
+- Fix bug in HDAWG8 driver where the gain was set to the range, whereas it should be gain = range/2 (#720)
```
QFer: Add this line to the 1.2.2 unreleased version changelog. The version 1.2.1 was already released.

PR #722, opened 2020-03-06
changes_review_generation_of_stepvalues_for_vector_scans

QFer:
* Reviewed code of an unreviewed change, did the adjustments
* Fixed bug in the functionality for 2d scanjob
* Cleaned code
* Boy scouting on file level
* Moved legacy code to legace
* Added unit test for scanjob 1d and 2d
* Fixed (errors/warning) with videomode integration test


src/qtt/measurements/scans.py
```diff
@@ -849,41 +850,43 @@ def _parse_2Dvec(self):
             self['stepdata'] = stepdata
             self['sweepdata'] = sweepdata
 
-    def _convert_scanjob_vec(self, station, sweeplength=None, steplength=None, stepvalues=None):
-        """ Adjust the scanjob for vector scans. 
+    def _convert_scanjob_vec(self, station, steplength=None, sweeplength=None, stepvalues=None) -> \
+            Tuple[SweepFixedValues, SweepFixedValues]:
```
peendebak: This violation our convention to either type annotation everything or nothing at all in a single fine. But I am fine with updating our policy to allow gradual type annotation of the code.

peendebak:
@sldesnoo-Delft @QFer Thanks for cleaning up the code! There is one thing that needs some attention: the PR is not backwards compatible, in the sense that now for the stepvalues for a scanjob are inclusive by default. For example: 
```
from qcodes import Parameter
from qtt.measurements.scans import scanjob_t
p = Parameter('p', set_cmd=None)
station=qcodes.Station()
scanjob = scanjob_t({'scantype': 'scan1Dfast', 'sweepdata': {'param': p, 'start': -2, 'end': 2, 'step': 1}})
_, sweepvalues = scanjob._convert_scanjob_vec(station)

print(list(sweepvalues))
print(np.arange(-2., 2, 1))
```
has output
```
[-2.0, -1.0, 0.0, 1.0, 2.0]
[-2. -1.  0.  1.]
```
Before the PR the output was 
```
[-2, -1, 0, 1]
[-2. -1.  0.  1.]
```` 
The old convention matches the convention of the python `range` and `np.arange` methods, so I would prefer that convention.

File: src/qtt/measurements/videomode.py [hunk 2]
```diff
@@ -434,6 +442,20 @@ def stop_all_instances():
...
+        """ Stop all VideoMode instances and cleanup """
+        global pgApp
+        lst = qtt.pgeometry.list_objects(VideoMode)
+        for v in lst:
+            v.stopreadout()
+            v.stop()
+            v.close()
+        pyqtgraph.cleanup()
+        if pgApp is not None:
+            del pgApp
```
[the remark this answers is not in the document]
QFer: Ok, I'll remove that part.

src/qtt/measurements/scans.py
QFer: I added this because it confused me that the return value was sometimes a list and sometimes one instance of SweepFixedValues. I changed that to always returning a tuple. Also I changed the mixture of slicing and sweepfixedvalues. By adding the typehint I wanted to be clear what to expect from this method and the fact that I changed the signature.

PR #723, opened 2020-03-20
Fix/sweep values

QFer:
* Adjusted _calculate_sweepvalues. It now always returns a valid SweepFixedValues object and doesn't raise an exception when the range is nog divisable by the stepvalue. The fix will adjust the end value do that the number of steps is fitting in the range.
* Added some extra unit test inclusive a test that fits random generated values for start, end and step.

### DIFF
--- a/src/qtt/measurements/scans.py
+++ b/src/qtt/measurements/scans.py
@@ -895,7 +895,7 @@ class scanjob_t(dict):
                 diff = abs(sweep_data['end'] - sweep_data['start'])
                 tolerance = 1e-10
                 if int(np.ceil(diff - tolerance)) == 0:
-                    raise ValueError('start and end values must differ at least 1 step value')
+                    raise ValueError('start and end values may not be the same')
 
                 sweep_values = param[sweep_data['start']:sweep_data['end']:sweep_data['step']]
 
--- a/src/qtt/measurements/scans.py
+++ b/src/qtt/measurements/scans.py
@@ -870,33 +870,41 @@ class scanjob_t(dict):
         """
 
         def _update_step_size_from_length(data, length):
-            """ From desired number of data points determine the required step size """
+            """ From desired number of data points determine the required step size.
+            length > 1: to get 'inclusive' data with length = 1, start and end values must be the same and step value
+            must be 0 which are invalid values.
+            """
             if length is not None:
                 if length == 1:
-                    data['step'] = 0
+                    raise ValueError('length must be > 1')
                 else:
                     if 'range' in sweepdata:
                         data['step'] = data['range'] / (length-1)
                     else:
                         data['step'] = (data['end'] - data['start']) / (length-1)
 
-        def _calculate_sweepvalues(param, sweep_data) -> SweepFixedValues:
-            """ From parameter and sweep data create a SweepFixedValues object. When (end - start) / step is not an
-            integer value, the end-value will be adjusted (lowered) until it is.
-             """
-            if sweep_data['step'] is not None:
-                steps = abs((sweep_data['end'] - sweep_data['start']) / sweep_data['step'])
-                tolerance = 1e-10
-                steps_lo = int(np.floor(steps + tolerance))
-                steps_hi = int(np.ceil(steps - tolerance))
+        def _calculate_sweepvalues(param, sweep_data, inclusive_end=False) -> SweepFixedValues:
+            """ From parameter and sweep data create a SweepFixedValues object. """
+            if sweep_data['step'] == 0.:
+                raise ValueError('step value may not be 0')
+            steps = abs((sweep_data['end'] - sweep_data['start']) / sweep_data['step'])
+            tolerance = 1e-10
+            steps_lo = int(np.floor(steps + tolerance))
+            steps_hi = int(np.ceil(steps - tolerance))
 
+            if inclusive_end:
                 if steps_lo != steps_hi:
                     sweep_data['end'] = sweep_data['start'] + steps_lo * sweep_data['step']
 
-            sweep_values = SweepFixedValues(param,
-                                            start=sweep_data['start'],
-                                            stop=sweep_data['end'],
-                                            step=sweep_data['step'])
+                sweep_values = SweepFixedValues(param,
+                                                start=sweep_data['start'],
+                                                stop=sweep_data['end'],
+                                                step=sweep_data['step'])
+            else:
+                if steps_lo == 0:
+                    raise ValueError('start and end values may not be the same')
+                sweep_values = param[sweep_data['start']:sweep_data['end']:sweep_data['step']]
+
             return sweep_values
 
         if self['scantype'][:6] == 'scan1D':
@@ -929,11 +937,12 @@ class scanjob_t(dict):
             if sweeplength is not None:
                 _update_step_size_from_length(sweepdata, sweeplength)
 
+            inclusive = sweeplength is not None
             if self['scantype'] in ['scan1Dvec', 'scan1Dfastvec']:
                 gates = station.gates
 
                 sweepdata['end'] = sweepdata['start'] + sweepdata['range']
-                sweepvalues = _calculate_sweepvalues(sweepparam, sweepdata)
+                sweepvalues = _calculate_sweepvalues(sweepparam, sweepdata, inclusive)
 
                 param_init = {param: gates.get(param)
                               for param in sweepdata['param']}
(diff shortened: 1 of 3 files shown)


src/qtt/measurements/scans.py
```diff
@@ -881,7 +881,18 @@ def _update_step_size_from_length(data, length):
                         data['step'] = (data['end'] - data['start']) / (length-1)
 
         def _calculate_sweepvalues(param, sweep_data) -> SweepFixedValues:
-            """ From parameter and sweep data create a SweepFixedValues object """
+            """ From parameter and sweep data create a SweepFixedValues object. When (end - start) / step is not an
+            integer value, the end-value will be adjusted (lowered) until it is.
+             """
+            if sweep_data['step'] is not None:
```
peendebak: Is the `if` needed? The `sweep_data` can be input at high level without a `step`, but at this stage it should always have a value right? If it is None, then the construction of SweepFixedValues will fail.

src/tests/unittests/measurements/test_scans.py
```diff
@@ -110,8 +111,8 @@ def test_scanjob_record_label(self):
         dataset = scan1D(station, scanjob, liveplotwindow=False, verbose=0)
         self.assertTrue(dataset.location.endswith(record_label))
 
-    @staticmethod
-    def test_scan2D(verbose=0):
+    def test_scan2D(self):
+        verbose = 0
```
peendebak: The `verbose=0` argument was used to help using and debugging the test functions. If it is removed as a default argument, we might as well remove it altogether.

src/tests/unittests/measurements/test_scans.py
```diff
@@ -364,6 +362,69 @@ def test_convert_scanjob_vec_scan1Dfast_adjust_sweeplength(self):
...
+        p = Parameter('p', set_cmd=None)
+        gates = VirtualIVVI(
+            name=qtt.measurements.scans.instrumentName('gates'), model=None)
+        station = qcodes.Station(gates)
+        station.gates = gates
+
+        scanjob = scanjob_t({'scantype': 'scan1Dfast',
+                             'sweepdata': {'param': p, 'start': 20, 'end': 20, 'step': .0075}})
+        _, sweepvalues = scanjob._convert_scanjob_vec(station)
+        actual_values = sweepvalues._values
```
peendebak:
```suggestion
        actual_values = list(sweepvalues)
```
Do not use the private `_values`. Altenatively, use `np.array(sweepvalues)`

src/qtt/measurements/scans.py
QFer: You are right, I will delete this test

src/tests/unittests/measurements/test_scans.py
QFer: This was a temporary quick fix, I accidentally left in. I will put it back.

src/tests/unittests/measurements/test_scans.py
QFer: This is a nice solution. I will change this.

QFer:
Your comment on this solution 'Not sure whether we get more issues because the default for a scanjob is now to be inclusive. Perhaps ask some people to test this now instead of before the next release.' is something I didn't want to happen. I repaired this. The functionality should not change. I added a few checks where an ValueError exception is raised with a message that is more descriptive than the error we got when it is generating an error later on  (for example divide by zero error).

src/qtt/measurements/scans.py
```diff
@@ -870,22 +870,41 @@ def _convert_scanjob_vec(self, station, steplength=None, sweeplength=None, stepv
...
+            if sweep_data['step'] == 0.:
+                raise ValueError('step value may not be 0')
+            steps = abs((sweep_data['end'] - sweep_data['start']) / sweep_data['step'])
+            tolerance = 1e-10
+            steps_lo = int(np.floor(steps + tolerance))
+            steps_hi = int(np.ceil(steps - tolerance))
+
+            if inclusive_end:
+                if steps_lo != steps_hi:
+                    sweep_data['end'] = sweep_data['start'] + steps_lo * sweep_data['step']
```
peendebak: This modifies the `sweep_data`. That is a bit shady, unless already a deepcopy was made before.

src/qtt/measurements/scans.py
```diff
@@ -870,22 +870,41 @@ def _convert_scanjob_vec(self, station, steplength=None, sweeplength=None, stepv
...
+
+                sweep_values = SweepFixedValues(param,
+                                                start=sweep_data['start'],
+                                                stop=sweep_data['end'],
```
peendebak:
```suggestion
                                                stop=sweep_data['end'] + .5*sweep_data['step'],
```
To avoid any rounding errors? Not sure whether they can occur

src/qtt/measurements/scans.py
sldesnoo-Delft: The dictionary sweep_data is passed during construction. So, any changes could have external effects.
sweep_data is also changed in many other places, e.g. lines 914, 915, 919,920, 924, 975, ...

src/qtt/measurements/scans.py
sldesnoo-Delft: Adding 0.5*step might not be a good idea, since qcodes determines the number of steps with the same code as lines 890-893 and raises an exception when steps_lo != steps_hi, a condition that is likely to occur when 0.5*step is added.

src/qtt/measurements/scans.py
```diff
@@ -870,22 +870,41 @@ def _convert_scanjob_vec(self, station, steplength=None, sweeplength=None, stepv
...
+                                                step=sweep_data['step'])
+            else:
+                if steps_hi == 0:
+                    raise ValueError('start and end values may not be the same')
```
sldesnoo-Delft, line 905: I would change the text of the error. It could be that the step is too big. I would say 'start and end values must differ at least 1 step value'

src/qtt/measurements/scans.py
```diff
@@ -870,22 +870,41 @@ def _convert_scanjob_vec(self, station, steplength=None, sweeplength=None, stepv
...
+                sweep_values = param[sweep_data['start']:sweep_data['end']:sweep_data['step']]
```
sldesnoo-Delft, line 906: Does a slice of `param` also return a SweepFixedValues object?

src/qtt/measurements/scans.py
QFer: You are right, this mutable parameter is adjusted in a lot of places. What is different is that this is an extra adjustment so in the end this new code has some side effect that was not in before.
I've thought about what I was trying to fix for this part of the original change but I think I am doing more bad than good. So what I will do is put back the old code for this 'inclusive end' part and only do the checks because these checks will lead to more understandable error messages.

src/qtt/measurements/scans.py
QFer: Yes, you are right Sander. With adjusting the 'end' parameter in the previous lines I assured that there was never a rounding error in the SweepFixValue creation.

src/qtt/measurements/scans.py
QFer: Nice suggestion.

src/qtt/measurements/scans.py
QFer: Yep. The __getitem__ of Parameter class is overridden to return the SweepFixedValue from a slice.

QFer:
The last few comments were about the generating the 'including_end' SweepFixedValues. I've thought about what I was trying to fix for this part of the original change but I think I am doing more bad than good. So what I will do is put back the old code for this 'inclusive end' part and only do the checks because these checks will lead to more understandable error messages.

src/qtt/measurements/scans.py
```diff
@@ -870,22 +870,35 @@ def _convert_scanjob_vec(self, station, steplength=None, sweeplength=None, stepv
...
+                raise ValueError('step value may not be 0')
+
+            if inclusive_end:
+                epsilon = 1e-8
+                sweep_values = param[sweep_data['start']:(sweep_data['end'] + epsilon):sweep_data['step']]
+            else:
+                diff = abs(sweep_data['end'] - sweep_data['start'])
+                tolerance = 1e-10
+                if int(np.ceil(diff - tolerance)) == 0:
+                    raise ValueError('start and end values must differ at least 1 step value')
```
peendebak, line 898: Here you check that start and end are not equal (upto the tolerance), but the error message is different. Do you want to check `diff > sweep_data['step']` or do you want to update the error message?

QFer: I changed the message because end-start doesn't need to be > step. I added a test for this situation. The text was from a previous situation with other circumstances. Good notice!

PR #725, opened 2020-03-20
[DEM-1368] initial investigation and destructor

bogdanvconstantin:
Adding a destructor to InstrumentDataClient that calls the ZmqRpcClient's ZmqSender destroy() method.


src/qtt/instrument_drivers/DistributedInstrument.py
```diff
@@ -42,6 +42,9 @@ def add_measurable_quantity(self, name='quantity', unit='arb.',
                           default_value=default_value, params=params)
         self.add_parameter(name, unit=unit, get_cmd=command,
                            docstring=doc_string)
+
```
oschusler: According to codacy, "the PR could be better". It's about trailing white space on this line

PR #729, opened 2020-03-27
use lmfit for fitting a gauss ramsey

peendebak:
* With the lmfit we get access to the reduced chi squared of the fit. Also the output format is in the default format now.
* In the previous implementation the first data point would be discarded in the fit. This has been fixed.

### DIFF
--- a/src/qtt/algorithms/functions.py
+++ b/src/qtt/algorithms/functions.py
@@ -11,7 +11,6 @@ import qtt.utilities.tools
 from qtt.utilities.visualization import plot_vertical_line
 from qtt.algorithms.generic import subpixelmax
 
-
 def gaussian(x, mean, std, amplitude=1, offset=0):
     """ Model for Gaussian function
 
@@ -337,20 +336,33 @@ def fit_gauss_ramsey(x_data, y_data, weight_power=0, maxiter=None, maxfun=5000, 
 
     """
 
-    def func(params): return cost_gauss_ramsey(x_data, y_data, params, weight_power=weight_power)
+    def gauss_ramsey_model(x, amplitude, decay_time, frequency, phase, offset):
+            """  """
+            y = gauss_ramsey(x, [amplitude, decay_time, frequency, phase, offset])
+            return y
+
+    if weight_power !=0:
+        raise NotImplementedError('weight_power !=0 not implemented' )
 
     if initial_params is None:
-        initial_params = estimate_parameters_damped_sine_wave(x_data, y_data, exponent=2)
+        initial_parameters = estimate_parameters_damped_sine_wave(x_data, y_data, exponent=2)
+    else:
+       initial_parameters=initial_params
+    lmfit_model = Model(gauss_ramsey_model)
+    lmfit_model.set_param_hint('amplitude', min=0)
+    lmfit_result = lmfit_model.fit(
+        y_data, x=x_data, **dict(zip(lmfit_model.param_names, initial_parameters)), verbose=verbose>=2)
+    import qtt.algorithms.fitting.extract_lmfit_parameters
+    result_dict = qtt.algorithms.fitting.extract_lmfit_parameters(lmfit_model, lmfit_result)
 
-    fit_parameters = scipy.optimize.fmin(func, initial_params, maxiter=maxiter, maxfun=maxfun, disp=verbose >= 2)
+    result_dict['description']= 'Function to analyse the results of a Ramsey experiment, fitted function: gauss_ramsey = ' \
+                       'A * exp(-(x_data/t2s)**2) * sin(2pi*ramseyfreq * x_data - angle) +B'
 
-    result_dict = {
-        'description': 'Function to analyse the results of a Ramsey experiment, fitted function: gauss_ramsey = '
-                       'A * exp(-(x_data/t2s)**2) * sin(2pi*ramseyfreq * x_data - angle) +B',
-        'parameters fit': fit_parameters,
-        'parameters initial guess': initial_params}
+    # backwards compatibility
+    result_dict['parameters fit']=result_dict['fitted_parameters']
+    result_dict['parameters initial guess']=initial_parameters
 
-    return fit_parameters, result_dict
+    return result_dict['fitted_parameters'], result_dict
 
 
 def plot_gauss_ramsey_fit(x_data, y_data, fit_parameters, fig):
@@ -405,7 +417,7 @@ def FermiLinear(x, a, b, cc, A, T, l=1.16):
     r""" Fermi distribution with linear function added
 
     Arguments:
-        x (numpy array): independent variable 
+        x (numpy array): independent variable
         a, b (float): coefficients of linear part
         cc (float): center of Fermi distribution
         A (float): amplitude of Fermi distribution
--- a/src/qtt/algorithms/functions.py
+++ b/src/qtt/algorithms/functions.py
@@ -348,8 +348,7 @@ def fit_gauss_ramsey(x_data, y_data, weight_power=None, maxiter=None, maxfun=500
         weights = None
     else:
         diff_x = np.diff(x_data)
-        weights =  np.hstack( (diff_x[0], diff_x) )  ** weight_power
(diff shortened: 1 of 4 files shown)


src/qtt/algorithms/functions.py
```diff
@@ -337,20 +339,38 @@ def fit_gauss_ramsey(x_data, y_data, weight_power=0, maxiter=None, maxfun=5000,
...
+    def gauss_ramsey_model(x, amplitude, decay_time, frequency, phase, offset):
+        """  """
+        y = gauss_ramsey(x, [amplitude, decay_time, frequency, phase, offset])
+        return y
+
+    if weight_power is None:
+        weights = None
+    else:
+        diff_x = np.diff(x_data)
+        weights =  np.hstack( (diff_x[0], diff_x) )  ** weight_power
```
QFer, line 351: Too many spaces. I understand you put in some spaces to indicate we are dealing with a tuple instead of 2 parameters. Nice.
```suggestion
        weights = np.hstack( (diff_x[0], diff_x) ) ** weight_power
```

src/qtt/algorithms/functions.py
```diff
@@ -337,20 +339,38 @@ def fit_gauss_ramsey(x_data, y_data, weight_power=0, maxiter=None, maxfun=5000,
...
-        initial_params = estimate_parameters_damped_sine_wave(x_data, y_data, exponent=2)
+        initial_parameters = estimate_parameters_damped_sine_wave(x_data, y_data, exponent=2)
+    else:
+        initial_parameters = initial_params
+    lmfit_model = Model(gauss_ramsey_model)
+    lmfit_model.set_param_hint('amplitude', min=0)
+    lmfit_result = lmfit_model.fit(
+        y_data, x=x_data, **dict(zip(lmfit_model.param_names, initial_parameters)), verbose=verbose >= 2, weights = weights)
+
+    import qtt.algorithms.fitting
```
QFer, line 363: Why local import instead of global
`from qtt.algorithms.fitting import extract_lmfit_parameters`
We prefer the global one, unless there is something against it.

src/qtt/algorithms/functions.py
```diff
@@ -337,20 +339,38 @@ def fit_gauss_ramsey(x_data, y_data, weight_power=0, maxiter=None, maxfun=5000,
...
+    result_dict = qtt.algorithms.fitting.extract_lmfit_parameters(lmfit_model, lmfit_result)
 
-    fit_parameters = scipy.optimize.fmin(func, initial_params, maxiter=maxiter, maxfun=maxfun, disp=verbose >= 2)
+    result_dict['description'] = 'Function to analyse the results of a Ramsey experiment, fitted function: gauss_ramsey = ' \
+        'A * exp(-(x_data/t2s)**2) * sin(2pi*ramseyfreq * x_data - angle) +B'
```
QFer, line 367:
```suggestion
        'A * exp(-(x_data / t2s) **2) * sin(2 * pi * ramseyfreq * x_data - angle) + B'
```

src/qtt/algorithms/functions.py
```diff
@@ -337,20 +339,38 @@ def fit_gauss_ramsey(x_data, y_data, weight_power=0, maxiter=None, maxfun=5000,
...
 
-    result_dict = {
-        'description': 'Function to analyse the results of a Ramsey experiment, fitted function: gauss_ramsey = '
-                       'A * exp(-(x_data/t2s)**2) * sin(2pi*ramseyfreq * x_data - angle) +B',
-        'parameters fit': fit_parameters,
-        'parameters initial guess': initial_params}
+    # backwards compatibility
+    result_dict['parameters fit'] = result_dict['fitted_parameters']
```
QFer, line 370: Are you sure parameters fit is without connecting  underscore?

src/qtt/algorithms/functions.py
```diff
@@ -337,20 +339,38 @@ def fit_gauss_ramsey(x_data, y_data, weight_power=0, maxiter=None, maxfun=5000,
...
+    result_dict['parameters initial guess'] = initial_parameters
```
QFer, line 371: Same here connecting  underscores?

src/qtt/algorithms/functions.py
QFer: Somewhere else I found 'parameters fitted gaussian' as key value. I'm confused...

src/tests/unittests/algorithms/test_functions.py
```diff
@@ -150,10 +150,28 @@ def test_fit_gauss_ramsey(self, fig=None):
...
         par_fit_test, results = fit_gauss_ramsey(x_data, y_data)
 
-        self.assertAlmostEqual(par_fit_test[0], 0.33, places=1)
-        self.assertAlmostEqual(par_fit_test[1] * 1e6, 3.36, places=1)
+        self.assertAlmostEqual(par_fit_test[0], 0.303, places=1)
+        self.assertAlmostEqual(par_fit_test[1] * 1e6, 3.55, places=1)
         self.assertAlmostEqual(par_fit_test[2] * 1e-6, 0.36, places=1)
 
+    def test_fit_gauss_ramsey_weights(self, fig=None):
+        x_data = np.hstack( (np.linspace(0, 3*np.pi, 20), np.linspace(3*np.pi, 4*np.pi, 100) ) )
```
QFer, line 158: Maybe autopep this test.
I didn't get the plot, what can be the cause for that?

src/qtt/algorithms/functions.py
peendebak: This one I just copied, but I will improve it with part of the suggestions.

src/qtt/algorithms/functions.py
peendebak: The end goal is for all fitting methods to use the lmfit framework, and have all fitting methods use the sames names. For that there is a generic method: https://github.com/QuTech-Delft/qtt/blob/dev/src/qtt/algorithms/fitting.py#L30

With the line above I maintain backwards compatibility by also providing the field with the old name.

@sldesnoo-Delft @QFer

src/qtt/algorithms/functions.py
peendebak: Same argument as previous comment.

src/tests/unittests/algorithms/test_functions.py
peendebak: The default argument `fig=None` means no plotting. I modified the test to plot by default. 

(the reason it is `None` by default is because of the split of tests into unittest and integration test, there is a ticket to revert this split)

src/qtt/algorithms/functions.py
peendebak: The global import causes a circular reference. Solving that would mean moving the method from `qtt.algorithms.functions` to `qtt.algorithms.fitting` which is not backwards compatible.

src/qtt/algorithms/functions.py
QFer: I understand but I mean the backwards compatibiliy key here. Here you use 'parameters fit' but in https://github.com/QuTech-Delft/qtt/blob/2854e1ddedf4c10e471b5c69d4e246efd8d36ed7/src/qtt/algorithms/fitting.py#L247
the key 'parameters fitted gaussian' is used for storing the 'fitted_parameters'.
Are you sure this is the right backwards compatible key?

src/tests/unittests/algorithms/test_functions.py
QFer: No I didn't mean that, sorry for the misunderstanding. I still think unit test should not show any output that is not part of the test.
Even when I add a number to fig, No plot is shown in my environment? But I think this shouldn't be part of the review comment because it is likely something in my environment.

src/qtt/algorithms/functions.py
peendebak: This is a change of the `fit_gauss_ramsey`, the current keys are: https://github.com/QuTech-Delft/qtt/blob/dev/src/qtt/algorithms/functions.py#L350. So I think they are right. @QFer

PR #734, opened 2020-04-10
Update to sine fitting

peendebak:
* Fix docstring of sine
* Write method in usual order so that it reads easier for users
* Add option to make sure amplitude of fit is positive

@bogdanvconstantin @QFer


src/qtt/algorithms/fitting.py
```diff
@@ -255,20 +255,23 @@ def gaussian_model(x, mean, sigma, amplitude):  # type: ignore
     return result_dict['fitted_parameters'], result_dict
 
 
-def fit_sine(x_data: np.ndarray, y_data: np.ndarray, initial_parameters=None) -> Tuple[Dict[str, Any], Dict[str, Any]]:
+def fit_sine(x_data: np.ndarray, y_data: np.ndarray, initial_parameters=None, positive_amplitude =  True) -> Tuple[Dict[str, Any], Dict[str, Any]]:
```
bogdanvconstantin, line 258: no spaces before and after "="

bogdanvconstantin: also line too long

peendebak:
@bogdanvconstantin Comments addressed

PR #735, opened 2020-04-21
Update qtt to qcodes 0.13.0

peendebak:
* Update qtt to qcodes 0.13.0 by creating a `qtt.DataSet` and `qtt.DataArray`
* Maintain backwards compatibility with 0.12.x (by adding a qcodes 0.12 test to travis)

Using `qtt.DataSet` allows us to modify the imports when `qcodes` drops support for the `DataSet` altogether.

@QFer @sldesnoo-Delft


peendebak:
> 1. Why all these changes. I the release notes is explained how to make use of the legacy code by setting import_legacy_api in the config
> 2. When you have a reason not to use the import_legacy_api, I think importing the DataSet and DataArray in qtt is making it harder to understand. Maybe a quick fix but making it harder to maintain. Why not import all the legacy stuff from their legacy modules. Now you mix it all (qtt DataSet import and qcodes.data.data_set imports)

1. When users update qtt this will work, even without updating a config file
2. At some point I expect the `qcodes.data.data_set.DataSet` to be removed. At that point we can just point `qtt.DataSet` to some other location.

QFer:
> 
> 
> > 1. Why all these changes. I the release notes is explained how to make use of the legacy code by setting import_legacy_api in the config
> > 2. When you have a reason not to use the import_legacy_api, I think importing the DataSet and DataArray in qtt is making it harder to understand. Maybe a quick fix but making it harder to maintain. Why not import all the legacy stuff from their legacy modules. Now you mix it all (qtt DataSet import and qcodes.data.data_set imports)
> 
>     1. When users update qtt this will work, even without updating a config file
> 
>     2. At some point I expect the `qcodes.data.data_set.DataSet` to be removed. At that point we can just point `qtt.DataSet` to some other location.

That will not work because of my second point. Besides DataSet and DataAray there are a handful of other deprecated classes (DiskIO) and methods (load_data, new_data) we still import by qcodes module name. Those will break then.

.travis.yml
```diff
@@ -27,9 +27,20 @@ matrix:
     env: FULL_INSTALL=1
     services:
     - xvfb
+  - name: QCoDeS 0.12 installation
```
QFer, line 30: This part is not needed anymore. We use the QCodes stuff as usual.

src/qtt/data.py
```diff
@@ -67,7 +67,7 @@ def _dictionary_to_data_array(array_dictionary):
     array_full_name = array_dictionary['full_name']
     if array_full_name is None:
         array_full_name = array_name
-    data_array = qcodes.DataArray(name=array_name,
+    data_array = DataArray(name=array_name,
                                   full_name=array_dictionary['full_name'],
```
QFer, line 71: indentation changed for few lines

src/qtt/measurements/scans.py
```diff
@@ -2310,23 +2311,21 @@ def plotData(alldata, diff_dir=None, fig=1):
         imx = qtt.utilities.tools.diffImageSmooth(alldata.measured.ndarray, dy=diff_dir)
         name = 'diff_dir_%s' % diff_dir
         name = uniqueArrayName(alldata, name)
-        data_arr = qcodes.DataArray(name=name, label=name, array_id=name,
+        data_arr = DataArray(name=name, label=name, array_id=name,
```
QFer, line 2314: You didn't correct the next line(s) for the changed indentation.

peendebak: Did an autopep on the file.

PR #738, opened 2020-04-29
update installation notes to use virtualenv

peendebak:
On Windows the qtt examples fail if installed with venv instead of virtualenv, see https://github.com/pyqtgraph/pyqtgraph/issues/1052


peendebak:
> Did you test this solution. My pyCharm also uses virtualenv and in Python 3.7 it fails.

Yes, it Worked For Me (tm). But that was with manual commands, not with pycharm

PR #739, opened 2020-05-12
[DEM-NNN] add virtual awg settings to snapshot

peendebak:
The `VirtualAwg` contains a sub-instrument for the `awg_map` and some other data. This data is not includes in the snapshot of the `VirtualAwg`. This PR makes sure the data is added to the snapshot by creating a read-only parameter that just reads the snapshot from the sub-instrument.


src/qtt/instrument_drivers/virtualAwg/virtual_awg.py
```diff
@@ -59,6 +59,9 @@ def __init__(self, awgs=None, settings=None, name='virtual_awg', logger=logging,
         self._latest_sequence_data = {}
         self.enable_debug = False
 
+        self.settings.snapshot(update=True)
+        self.add_parameter('settings_snapshot', get_cmd = self.settings.snapshot, label='Settings snapshot')
```
QFer, line 63:
```suggestion
        self.add_parameter('settings_snapshot', get_cmd=self.settings.snapshot, label='Settings snapshot')
```
Some spaces.

QFer: Is it possible to add a unit test that makes clear why this functionality is added? When is the settings snapshot updated with the get_cmd?

peendebak: @QFer Added a test, and fixed the bug that was exposed.

PR #744, opened 2020-05-22
[OP-56] Replace polyintersect inplementation of Polygon3 with shapely

peendebak:
The depencency on `Polygon3` was hard to maintain over different python versions. `shapely` is better supported, at the cost of a slower implementation

### DIFF
--- a/src/qtt/pgeometry.py
+++ b/src/qtt/pgeometry.py
@@ -645,6 +645,7 @@ def directionMean(vec):
 
     """
     vec = np.array(vec)
+
     def dist(a, vec):
         phi = np.arctan2(vec[:, 0], vec[:, 1])
         x = a - phi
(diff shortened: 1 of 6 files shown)


src/qtt/pgeometry.py
```diff
@@ -32,15 +32,14 @@
...
+import numpy
 import numpy as np
-import Polygon as polygon3
 import scipy.io
-import numpy
 import scipy.ndimage.morphology as morphology
 import scipy.ndimage.filters as filters
-
-from functools import wraps
+import shapely
```
QFer, line 42:
```suggestion
import shapely.geometry
```

src/qtt/pgeometry.py
```diff
@@ -1136,32 +1135,32 @@ def polysegments(p):
     return 0.5 * abs(sum(x0 * y1 - x1 * y0 for ((x0, y0), (x1, y1)) in polysegments(p)))
 
 
-def polyintersect(x1, x2):
-    """ Intersection of two polygons
+def polyintersect(x1 : np.ndarray, x2 : np.ndarray) -> np.ndarray:
+        """ Calcualte intersection of two polygons
```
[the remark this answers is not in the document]
peendebak: Fixed. The autopep updates some more lines

peendebak:
> in test_pgeometry an error of Polygon is imported which doesn't exists anymore.
> The unit tests are failing. shapely works with closed polygons begin and endpoint (are the same but 2 points) while polygon didn't

The unit tests have been updates. The intersection polygon still has the same begin and endpoint. This is a valid definition of a polygon, and conforms better to the shapely convention.

src/qtt/pgeometry.py
```diff
@@ -1136,33 +1136,38 @@ def polysegments(p):
...
-    x = np.array(p)
-    x = x.reshape((-1, 2))
+
+    p1 = shapely.geometry.Polygon(x1)
+    p2 = shapely.geometry.Polygon(x2)
+    p = p1.intersection(p2)
+    if p.is_empty:
+        return np.zeros((0, 2))
+    if isinstance(p, shapely.geometry.multipolygon.MultiPolygon):
+        raise Exception('intersection of polygons is not a simple polygon')
```
QFer, line 1166: I would rather see a ValueError raised instead of the too generic Exception

src/tests/unittests/test_pgeometry.py
```diff
@@ -89,11 +89,26 @@ def test_polyintersect_degenerate_polygon(self):
...
-        except PolygonError as ex: # the assertRaises does not work due to the ctypes error message
-            self.assertIn('Invalid polygon or contour for operation', str(ex))
+
+    def test_multi_polygon_intersection(self):
+        delta = .42
+        shift = 1.7
+        x1 = np.array([(-1, -1), (1, -1), (1, 1), (-1, 1)]) + np.array([shift, 0])
+        x2 = np.array([(delta, 0), (5, 5), (-5, 5), (-delta, 0), (-5, -5), (5, -5), (delta, 0)])
+
+        with self.assertRaises(Exception):
```
QFer, line 101: When you did change it to ValueError also here we can do assertRaises(ValueError):
```suggestion
        with self.assertRaises(ValueError):
```

src/qtt/pgeometry.py
QFer, line 1166:
```suggestion
        raise ValueError('intersection of polygons is not a simple polygon')
```

PR #745, opened 2020-05-23
Fix Python 3.8 warnings

peendebak:
Make qtt compatible with python 3.8


QFer:
#744 should be fixed first

peendebak:
> If you rebase I approve

Done!

PR #746, opened 2020-06-04
[OP-NNN] QCoDes 0.14.0 compatibility

QFer:
* Fix for incompatibility with latest QCodes (0.14.0)

### DIFF
--- a/src/qtt/__init__.py
+++ b/src/qtt/__init__.py
@@ -170,7 +170,7 @@ def _copy_to_str(x, memo=None):
 
 
 def _setstate(self, d):
-    self.name = d
+    self._short_name = d
     self._instrument = None
 
     def _get():
--- a/src/tests/unittests/instrument_drivers/virtualAwg/test_virtual_awg.py
+++ b/src/tests/unittests/instrument_drivers/virtualAwg/test_virtual_awg.py
@@ -1,7 +1,6 @@
-from typing import List
+from typing import List, Any
 import unittest
 
-from qcodes import Instrument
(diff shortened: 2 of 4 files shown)


src/tests/unittests/instrument_drivers/virtualAwg/test_virtual_awg.py
```diff
@@ -15,7 +16,7 @@ def tearDown(self) -> None:
         self.settings.close()
 
     @staticmethod
-    def __create_awg_drivers() -> List[Mock]:
+    def __create_awg_drivers():
```
peendebak, line 19: Why remove the type annotation? If there are warnings of errors, can we not relax to something like `List[Any]`?

src/qtt/__init__.py
```diff
@@ -170,7 +170,7 @@ def _copy_to_str(x, memo=None):
 
 
 def _setstate(self, d):
-    self.name = d
+    self._short_name = d
```
peendebak, line 173: This does change the functionality, but I think it is ok since I doubt that we are using the serialization of Parameters with pickle.

src/tests/unittests/instrument_drivers/virtualAwg/test_virtual_awg.py
QFer: Type hinting unit tests is asking for mypy errors because in the unit tests we use Mock classes when calling production code instead of the production classes. But indeed, List[Any] is possible here.

PR #747, opened 2020-08-11
fix for numpy deprecation

peendebak:
A minimal example is:
```
import qtt
dataset = qtt.data.load_example_dataset('frequency_rabi_scan.json')
qtt.dataset_processing._slice_dataset(dataset, [slice(0, 4)], output_parameter_name=None, copy_metadata=True)
```
The warning message generated is
```
c:\projects\qcodes\qcodes\data\data_array.py:344: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  return self.ndarray[loop_indices]
```


bogdanvconstantin (approved):
Ok!

PR #748, opened 2020-09-04
[DEM-NNN] Fix ppt notes size

peendebak:
If notes are added to powerpoint which are too large powerpoint will freeze (has to be restarted to work again). This PR adds a safe limit to the size of the notes. The limit is configurable.

### DIFF
--- a/src/qtt/utilities/tools.py
+++ b/src/qtt/utilities/tools.py
@@ -853,7 +853,7 @@ try:
     def addPPTslide(title=None, fig=None, txt=None, notes=None, figsize=None,
                     subtitle=None, maintext=None, show=False, verbose=1,
                     activate_slide=True, ppLayout=None, extranotes=None, background_color=None,
-                    maximum_notes_size: int = 10e3):
+                    maximum_notes_size: int = 10000):
         """ Add slide to current active Powerpoint presentation.
 
         Arguments:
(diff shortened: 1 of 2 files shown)


src/qtt/utilities/tools.py
```diff
@@ -1142,7 +1148,6 @@ def addPPTslide(title=None, fig=None, txt=None, notes=None, figsize=None,
         warnings.warn(
             'addPPTslide is not available on your system. Install win32com from https://pypi.org/project/pypiwin32/.')
```
bogdanvconstantin, line 1149: Python (mypy) is confused about which addPPTslide function to use.

bogdanvconstantin (commented):
Will approve but pipeline needs to pass

src/qtt/utilities/tools.py
QFer: @bogdanvconstantin The problem here is that it is a conditional function (it is in a try except ImportError construction) depending on if you are running windows, which you aren't running. The parameters of both functions should at least be the same, which they aren't currently. mypy is complaining about that.

@peendebak you should have noticed the build failed (and fix of course) before asking for a review.

src/qtt/utilities/tools.py
```diff
@@ -850,10 +850,10 @@ def figure_ppt_callback(event):
     import win32com
     import win32com.client
 
-
     def addPPTslide(title=None, fig=None, txt=None, notes=None, figsize=None,
                     subtitle=None, maintext=None, show=False, verbose=1,
-                    activate_slide=True, ppLayout=None, extranotes=None, background_color=None):
+                    activate_slide=True, ppLayout=None, extranotes=None, background_color=None,
+                    maximum_notes_size: int = 10e3):
```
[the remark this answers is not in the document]
QFer: When you test it, it is implicitly 'converted' to 10000. Not what you want.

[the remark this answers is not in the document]
peendebak: 10000 is indeed quite some text, but i know that there are notes added of this size (because the snapshot is included) so I would not want to restrict more be default. 

A user can off course change the default value

src/qtt/utilities/tools.py
peendebak: Point taken

QFer (approved):

Issue #721, opened 2020-03-05
Add set of example data of qtt distribution


sldesnoo-Delft:
qtt should provide a set of example measurement data of common measurements. This data can be used as input during development and demonstration, also for code that is not part of qtt.

There is already some example data, but it is not included in the distribution and does not have descriptive names.

Changes to be made:
* Include example data in distribution
* Change names to descriptive ones
* Add a function to retrieve the names of the available examples

Rename:
"addition_line_scan" Old name: '17-57-24'
"charge_stability_diagram_anti_crossing" Old name: '2017-01-10/09-45-15'
"polarization_line" Old name: '2017-02-21/15-59-56'
"charge_stability_diagram_anti_crossing_virtual_gates" Old name: '2017-09-04/11-04-14_qtt_scan2Dturbovec'
"PAT_scan_background" Old name: '2017-09-04/11-04-45_qtt_scan1Dfastvec'
"PAT_scan" Old name: '2017-09-04/11-05-17_qtt_scan2Dfastvec'
"coulomb_peak" Old name: 'misc/13-13-16_qtt_sensingdot_t.fastTune'
"charge_stability_diagram_double_dot_system" Old name: 'misc/Lever_arm_Charging_energy/10-06-59_qtt_scan2Dfast' (and perhaps resample to make loading faster)
"charge_stability_diagram_double_dot_system_detail" Old name: 'misc/Lever_arm_Charging_energy/14-28-52_qtt_scan2Dfast'
"charge_stability_diagram_double_dot_system_bias_triangle" Old name: 'misc/Lever_arm_Charging_energy/14-33-26_qtt_scan2Dfast'
"pinchoff_scan_barrier_gate" Old name: 'misc/pinchoff/2018-06-28/13-24-22_qtt_scan1D'
"2electron_pat_background" Old name: 'misc/2electron_pat_26102017/background'
"2electron_pat" Old name: 'misc/2electron_pat_26102017/pat'
"elzerman_detuning_scan" Old name: 'elzerman_detuning_scan/2019-05-28_11-39-39_qtt_generic.json'
'charge_stability_diagram_dac_vs_awg' Old name 'misc/P5_y_P5_x.dat'
'rts_signal' Old name '2018-08-31/19-05-18_qtt_save_segments'

Name already good:
'time_rabi_scan.json'
'frequency_rabi_scan.json'
'double_gaussian_dataset.json'

To be removed from repository:
'misc/awg_to_plunger_scan.pickle'

[...]

Issue #715, opened 2020-02-23
exampledata folder seems to not be cloned


bogdanvconstantin:
I installed qtt as a site-package part of another package.

In my setup.py I have:

```
qtt = 'qtt @ git+ssh://git@github.com/QuTech-Delft/qtt#egg=qtt'
...
setup(
    ...
    install_requires = [ ..., qtt, ...]
    ...
    )
```

However, it seems that after the installation completes successfully (after a `pip install .`)the qtt/exampledata/ folder is not part of my qtt site-packages. 

![Uploading Screenshot 2020-02-23 at 20.08.09.png…]()

Can someone please have a look at this? 
Thank you!

peendebak:
The exampledata should be accesible with 
```
import qtt.data
dataset=qtt.data.load_example_dataset('time_rabi_scan.json')
```
The `load_example_dataset` is used in several example notebooks, e.g. https://qtt.readthedocs.io/en/latest/notebooks/analysis/example_fermi_fitting.html
@sldesnoo-Delft @QFer

QFer:
When installing qtt via 'pip install .' only the qtt 'functionality' is installed, not the docs, ipython notebooks etc. So in my opinion exampledata is also one of the components that should not be installed. When you want to run examples or unit tests you have to install in editable mode.

Or am I missing something?

sldesnoo-Delft:
@QFer You've got a good point. I tend to agree and we should remove the docs and examples from the install.
@peendebak Do you agree?

sldesnoo-Delft:
See #721

QFer:
Closed because this issue is solved in https://github.com/QuTech-Delft/qtt/pull/750

PR #750, opened 2020-11-30
[OP-111] Example data renaming/installing and unit test fixes

QFer:
The PR consists a solution for:
* #715 exampledata folder seems to not be cloned
* #719 include sampledata using Manifest
* #721 Add set of example data of qtt distribution
* #749 Bump notebook from 6.0.3 to 6.1.5
* Fixing unit test on version 1.2.2


src/tests/unittests/test_dataset_processing.py
```diff
@@ -9,11 +9,11 @@
 class TestDataProcessing(unittest.TestCase):
 
     def test_slice_dataset(self):
-        dataset = qcodes.tests.data_mocks.DataSet1D()
+        dataset = qcodes.tests.legacy.data_mocks.DataSet1D()
```
eendebakpt, line 12: Since we use this method (and the `DataSet2D`) a lot and it is legacy in qcodes, maybe import `qcodes.tests.legacy.data_mocks.DataSet1D` into `qtt.data` and use `qtt.data.DataSet1D` everywhere?

(we can also ignore this for now and do it once it comes up again)

CHANGELOG.md
```diff
@@ -7,11 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Dev
 
 ### Added
-- Support for python 3.8 (#672)
+- Support for python 3.8 added (#672)
 
 ### Changed
 - Replaced implementation of polyintersection from Polygon3 to shapely (#744)
 
+### Removed
```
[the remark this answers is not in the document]
QFer: I will update the changelog with all the changes

src/tests/unittests/test_dataset_processing.py
QFer: I don't want a dependency between QTT production code (qtt.data) and QCoDeS test code. As an alternative I do an explicit import of DataSet1D and DataSet2D where they are used in our unit tests. In the tests we use DataSet1D instead of using the legacy path everywhere.

eendebakpt:
That is even better!

eendebakpt (approved):

PR #753, opened 2021-03-09
[OP-NN] Fix warnings in unit tests

QFer:
* Fix warnings in unit tests
* Fix a vulnerability in bleach
* Updated requirements_lock.txt with the latest dependencies
* Adjusted changelog
* Moved some tests with windows to integration (because they crash while running in linux)
* Added a version.py in the root to conform to the standard


src/tests/unittests/test_structures.py
```diff
@@ -34,7 +34,8 @@ def test_sensingdot_t(fig=None, verbose=0):
         sensing_dot = qtt.structures.sensingdot_t(
             ['SD1a', 'SD1b', 'SD1c'], station=station, minstrument='keithley1.amplitude')
         sensing_dot.verbose = verbose
-        _ = sensing_dot.autoTune(step=-8, fig=fig)
+        if fig is not None:
+            _ = sensing_dot.autoTune(step=-8, fig=fig)
```
peendebak, line 38: `autoTune` does more than plotting. So it should be executed in the unittest

version.py
```diff
@@ -0,0 +1,5 @@
+if __name__ == '__main__':
```
peendebak, line 1: Why add this file?

peendebak (commented):
A few small items. I also created #755 to be merged into this branch

src/qtt/data.py
```diff
@@ -1178,7 +1178,12 @@ def makeDataSet1Dplain(xname, x, yname, y=None, xunit=None, yunit=None, location
     setpoint_data = np.array(x)
     preset_data = np.NaN * np.ones(setpoint_data.size)
     if y is not None:
-        y = np.array(y)
+        if isinstance(y, np.ndarray):
+            y = np.array(y, dtype=y.dtype)
+        elif isinstance(y, list):
+            y = np.array(y, dtype=type(y[0]))
```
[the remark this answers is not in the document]
QFer: Fixed the empty y case. I don't think it is a valid case but the old code did 'accept' it with a warning although I don't know if the result is usable when there is a list without measured data. I added a test for this case.

It's not a mypy but a numpy deprecation warning.
```
VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify
'dtype=object' when creating the ndarray.
```

src/tests/unittests/test_structures.py
QFer: I copied the non-silent test to integration. Because all the windows popping up is crashing wildly on linux (python abort).
In the default pytest settings (pytest.ini) also the tests in directory integration are executed as unit test. On linux this pytest.ini can be adjusted to only test the 'silent' unit tests. When necessary we can add another pytest configuration file. One for integration tests and one for silent tests.

version.py
QFer: To be able to get the version of a package via a standard way, running this version.py in the root of the package. It can be used by other tools to check the version (even of a non-installed package) or get the version automatically when releasing to master.

src/qtt/algorithms/allxy.py
```diff
@@ -49,13 +49,13 @@ def allxy_model(indices: Union[float, np.ndarray], offset0: float, slope0: float
...
 
-def _estimate_allxy_parameters(allxy_data: np.ndarray) -> List[float]:
+def _estimate_allxy_parameters(allxy_data: np.ndarray) -> List[Any]:
     """ Return estimate of allxy model parameters """
     p = [np.mean(allxy_data[0:5]), 0, np.mean(allxy_data[5:17]), 0, np.mean(allxy_data[17:]), 0]
     return p
 
 
-def _default_measurement_array(dataset: DataSet) -> DataArray:
+def _default_measurement_array(dataset: DataSet) -> np.ndarray:
```
peendebak, line 58: The `_default_measurement_array` does return a `DataArray`. Annotating it as `np.ndarray` is not entirely correct, but ok since the DataArray converts to `np.ndarray` automatically.

The way to do it would be to use the `ArrayLike` type (see https://numpy.org/devdocs/reference/typing.html). This requires a more recent numpy though, so suggest to leave the annotation as it is.

peendebak (approved):

PR #756, opened 2021-03-22
[OP-NN] Release 1.2.3

QFer:
* Updated requirements_lock.txt
* Updated CHANGELOG
* Silenced a unit test


peendebak (approved):
PR itself is accepted. Perhaps first look at https://github.com/QuTech-Delft/qilib/pull/101 and release a new version if qilib? That prevents releasing qtt now and finding out issues with qilib when that is updated.

PR #764, opened 2021-06-02
Remove deprecated code

eendebakpt:
This PR removes code that has been deprecated (and throwing an error when used) for a long time.


src/qtt/simulation/dotsystem.py
```diff
@@ -270,9 +270,6 @@ def __init__(self, name='dotsystem', ndots=3, **kwargs):
         self.ndots = ndots
         self.temperature = 0
 
-        self.makevars = qtt.utilities.tools.deprecated(self.make_variables)
-        self.makevarsM = qtt.utilities.tools.deprecated(self._make_variable_matrices)
-
```
QFer, line 275: There are 2 more instances of @qtt.utilities.tools.deprecated in this module (getHn, defaultVmatrix). why are they not removed?

A search also shows instances of @qtt.utilities.tools.deprecated or @qtt.utilities.tools.rdeprecated in:
generic.py show2Dimage
data.py 2 instances
scans.py 2 instances
functions.py _cost_double_gaussian
instrument_storage.py, 2 instances
virtual_gates.py 1 instance

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- Removed deprecated code (#764)
+
```
QFer, line 18: In conf.py I see the following deprecated stuff in ignore_paths:

'../qtt/deprecated/.*.py', '../qtt/deprecated/tunnelbarrier.py'

the latter one is enclosed in the first one and this file tunnelbarrier.py doesn't exist anymore.

QFer (changes requested):
I would say when we get rid of the deprecated stuff in qtt, get rid of all deprecated functions/methods. I found a few more. I added them to the review comments

src/qtt/simulation/dotsystem.py
eendebakpt: The `rdeprecated` throws an exception, so we can be sure everyone using it should be aware of the deprecation. The `deprecated` only prints a warning, so I wanted to be on the safe side.
I will remove most `deprecated` code as well, or convert from `deprecated` to `rdeprecated`

QFer (approved):

PR #765, opened 2021-06-02
Add measure_time context

eendebakpt:


src/qtt/utilities/tools.py
```diff
@@ -1,71 +1,99 @@
...
     import qtpy.QtWidgets as QtWidgets
 except BaseException:
     pass
 
 
 # %%
 
-from typing import Optional
-import contextlib
+def measure_time(txt: str = 'dt:'):
```
QFer, line 67: Why the extra function returning the class? When you would define the class only you can do (from test_tools.py):
```
with measure_time_context('hi') as m:
    time.sleep(.1)
```
and get the same results

(when you agree you may consider to rename the class to measure_time to keep the rest the same)

src/tests/unittests/utilities/test_tools.py
```diff
@@ -1,19 +1,34 @@
...
 import numpy as np
 import qcodes
 from qcodes.data.data_array import DataArray
 from qcodes.tests.legacy.data_mocks import DataSet2D
-from qtt.utilities import tools
-from qtt.utilities.tools import resampleImage, diffImage, diffImageSmooth, reshape_metadata, get_python_version, \
-    get_module_versions, get_git_versions, code_version, rdeprecated, in_ipynb, pythonVersion
+
 import qtt.measurements.scans
+from qtt.utilities import tools
```
QFer, line 15: Do you want this if in the next line you import the single dependencies. You have to replace e.g. tools.code_version with code_version etc. and we get rid of the doubly imported stuff.

src/qtt/utilities/tools.py
```diff
@@ -971,7 +1000,8 @@ def addPPTslide(title=None, fig=None, txt=None, notes=None, figsize=None,
         else:
             slide.shapes.title.textframe.textrange.text = 'QCoDeS measurement'
 
-        from qtt.measurements.videomode import VideoMode  # import here, to prevent default imports of gui code
+        from qtt.measurements.videomode import \
```
QFer, line 1003: Why the line break? Line was not too long.
Why break the line here instead of sprint the comment from the code
```
# import here, to prevent default imports of gui code
from qtt.measurements.videomode import VideoMode

```

QFer (changes requested):

src/tests/unittests/utilities/test_tools.py
eendebakpt: Not part of the PR, but I removed it anyway

QFer (dismissed):

QFer (approved):

PR #766, opened 2021-06-02
add default value for fig argument of create_figure_ppt_callback

eendebakpt:


src/qtt/utilities/tools.py
```diff
@@ -814,21 +814,22 @@ def _ppt_determine_image_position(ppt, figsize, fname, verbose=1):
     return left, top, width, height
 
 
-def create_figure_ppt_callback(fig, title=None, notes=None, position=(0.9, 0.925, 0.075, 0.05)):
-    """ Create a callback on a matplotlib figure to copy data to PowerPoint slide.
+def create_figure_ppt_callback(fig : Optional[int] = None, title : Optional[str] = None,
+                       notes : Optional[Union[str, DataSet]]=None, position : Tuple =(0.9, 0.925, 0.075, 0.05)):
```
QFer, line 818: 1. Why not also add a return type
2. type Tuple is generic, it can be made more specific Tuple[float, float, float, float]
3. missing spaces surrounding the = between ]]=None and =(0.9
```suggestion
                               notes : Optional[Union[str, DataSet]] = None,
                               position : Tuple[float, float, float, float] = (0.9, 0.925, 0.075, 0.05)) -> None:
```

src/qtt/utilities/tools.py
```diff
@@ -814,21 +814,22 @@ def _ppt_determine_image_position(ppt, figsize, fname, verbose=1):
...
+    """ Create a button on a matplotlib figure to copy data to PowerPoint slide.
 
     The figure is copied to PowerPoint using @ref addPPTslide.
 
     Args:
-        fig (int): handle to matplotlib window.
-        title (None or str): title for the slide.
-        notes (None or str or DataSet): notes to add to the slide.
+        fig): Handle to matplotlib window. If None, then use the current figure
```
QFer, line 824: fig) -> fig

QFer (changes requested):

QFer (approved):

PR #776, opened 2021-07-08
add .pycodestyle file

eendebakpt:
Add a `.pycodestyle` file that replaces the argument in the pre-commit config. See https://github.com/pre-commit/mirrors-autopep8


QFer (commented):
When is this file read? I suppose you configure it on **project level**. https://pycodestyle.pycqa.org/en/latest/intro.html#configuration description doesn't describe reading .pycodestyle from the project root?

peendebak:
> When is this file read? I suppose you configure it on **project level**. https://pycodestyle.pycqa.org/en/latest/intro.html#configuration description doesn't describe reading .pycodestyle from the project root?

There are dozens of options available for configuring these tools. I updated so that it now takes one which might be a bit more general and is included in the docs. https://github.com/hhatto/autopep8#configuration

@QFer

QFer (approved):

PR #777, opened 2021-07-12
Deprecate legacy classes and interfaces

peendebak:
* Deprecate instrument adapters
* Deprecate measurement interfaces
* Deprecate post-processing code using qilib dataset

### DIFF
--- a/src/tests/unittests/measurements/processing/test_process_sawtooth_2d.py
+++ b/src/tests/unittests/measurements/processing/test_process_sawtooth_2d.py
@@ -1,8 +1,9 @@
+from unittest import TestCase
+
 import numpy as np
+from qilib.data_set import DataArray, DataSet
 from scipy.signal import sawtooth
 
-from unittest import TestCase
-from qilib.data_set import DataArray, DataSet
 from qtt.measurements.post_processing import ProcessSawtooth2D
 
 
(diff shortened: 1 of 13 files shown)


src/tests/unittests/measurements/processing/test_process_sawtooth_2d.py
```diff
@@ -62,8 +63,8 @@ def __dummy_scope_data(time_data, sawteeth_count, period, width):
 
     @staticmethod
     def __dummy_data_array(set_array: DataArray, scope_data: np.ndarray, channel_index: int = 1, trace_number: int = 1):
-        idenifier = 'ScopeTrace_{:03d}'.format(trace_number)
-        label = 'Channel_{}'.format(channel_index)
+        idenifier = f'ScopeTrace_{trace_number:03d}'
```
QFer, line 66: Idenifier->Identifier

QFer (commented):
Why are all these instrument adapters and post_processing deprecated, what do we use instead?

peendebak:
> Why are all these instrument adapters and post_processing deprecated, what do we use instead?

The adapters and post processing have not been used. We have always used the post processing in `qtt.measurements.scans`.

src/tests/unittests/measurements/processing/test_process_sawtooth_2d.py
peendebak: The spelling mistake was already in the original code and this is part of the deprecated class.

QFer (approved):

PR #791, opened 2021-09-27
add regression test and fix for gaussian fit

peendebak:
Add a case where the gaussian fitting would fail. This is fixed by changing the fitting method for lmfit.

### DIFF
--- a/src/qtt/algorithms/fitting.py
+++ b/src/qtt/algorithms/fitting.py
@@ -217,6 +217,9 @@ def fit_gaussian(x_data, y_data, maxiter=None, maxfun=None, verbose=0, initial_p
                  estimate_offset=True):
     """ Fitting of a gaussian, see function 'gaussian' for the model that is fitted
 
+    The final optimization of the fit is performed with `lmfit <https://lmfit.github.io/lmfit-py/>`
+    using the `least_squares` method.
+
     Args:
         x_data (array): x values of the data
         y_data (array): y values of the data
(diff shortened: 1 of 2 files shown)


src/qtt/algorithms/fitting.py
```diff
@@ -248,10 +249,11 @@ def gaussian_model(x, mean, sigma, amplitude):  # type: ignore
             y = gaussian(x, mean, sigma, amplitude)
             return y
 
+    lmfit_method = 'least_squares'
     lmfit_model = Model(gaussian_model)
     lmfit_model.set_param_hint('amplitude', min=0)
     lmfit_result = lmfit_model.fit(
-        y_data, x=x_data, **dict(zip(lmfit_model.param_names, initial_parameters)), verbose=verbose)
+        y_data, x=x_data, **dict(zip(lmfit_model.param_names, initial_parameters)), verbose=verbose, method=lmfit_method)
```
QFer, line 256: line too long

QFer, line 256: Maybe add to the docstring that 'least_squares' is used as the fitting-method instead of the default 'leastsq'

src/qtt/algorithms/fitting.py
```diff
@@ -248,10 +249,11 @@ def gaussian_model(x, mean, sigma, amplitude):  # type: ignore
             y = gaussian(x, mean, sigma, amplitude)
             return y
 
+    lmfit_method = 'least_squares'
```
QFer, line 252: Check:  Shouldn't fit method in fit_double_gaussian also change, because fit_gaussian and fit_double_gaussian are both used in refit_double_gaussian, each using a different minimizer now.

QFer (changes requested):

src/qtt/algorithms/fitting.py
peendebak: Updated the docstrings

src/qtt/algorithms/fitting.py
peendebak: I don't believe that is a problem, but I changed it anyway.

QFer (approved):

Issue #799, opened 2021-11-18
Problems with DoubleDot from classicaldotsystem


fabian-fuchs:
Hello,

I tried simulating a DoubleDot analog to the classical_triple_dot.ipynb example notebook.

```
%matplotlib inline
import matplotlib.pyplot as plt

import numpy as np
from qtt.simulation.classicaldotsystem import ClassicalDotSystem, DoubleDot

DotSystem = DoubleDot()

# addition energy
DotSystem.Eadd = np.array([54.0, 52.8]) 

# chemical potential at zero gate voltage
DotSystem.mu0 = np.array([120.0, 100.0])  

# coulomb repulsion (!order is important: (1,2), (1,3), (2,3)) (lexicographic ordering)
DotSystem.alpha = np.array([[1.0, 0.25], [0.25, 1.0]])
# virtual gate matrix, mapping gates to chemical potentials
DotSystem.W = np.array([6.0])

#Define gate plane
P1 = 25.987333
P2 = 83.0536667
ref_value = [P1, P2]
gatevalues = np.array(ref_value)
charge_state = DotSystem.calculate_ground_state(gatevalues)
```

But I get the following error:
![qtt_bug](https://user-images.githubusercontent.com/73223556/142383144-17765699-2a8a-4d12-8712-0d64189d2112.PNG)

I think the reason is that by creating an instance of  ```DoubleDot```  the method ```makebasis``` is called twice with different parameters. The first call is in the ```init``` method of  ```ClassicalDotSystem``` with the parmeters ```self.makebasis(ndots=ndots, maxelectrons=maxelectrons)```. And the second call is in the  ```init``` method of ```DoubleDot``` with the parmeters ```self.makebasis(ndots=2)```. The first call creates a numpy array with shape (16,2) from which the variable ```_coulomb_energy``` inherits it shape but the second call overrides the variable ```basis``` with a array of shape (9,2). Therefore the shapes between the parameters ```basis``` and ```_coulomb_energy``` don´t match anymore, which results in the error.

As a possible solution I added the parameter ```maxelectrons``` to the class ```DoubleDot``` analog to the class ```TripleDot```.
Code before:
```
class DoubleDot(ClassicalDotSystem):
    """ Classical simulation of double dot """

    def __init__(self, name='doubledot', **kwargs):
        super().__init__(name=name, ndots=2, ngates=2, **kwargs)

        self.makebasis(ndots=2)
```
Code after:
```
[...]

PR #801, opened 2021-11-23
[SDS-NNN] issue dotsystem makebasis

QFer:
* restructuring calls to makebasis, now it is only called (once) from base class
* The subclasses pass the necessary info through the constructor chain.
* Solves #799


peendebak (approved):
Looks good to me

PR #802, opened 2021-11-28
remove legacy construction for monitorSizes

eendebakpt:
* The `try - except BaseException:` is a legacy construction that is not used any more. It is removed
* Add type annotation for return value of `monitorSizes`

@QFer


src/qtt/pgeometry.py
```diff
@@ -1755,37 +1755,38 @@ def ff(xx, figx=fig): return mpl2clipboard(fig=figx)
...
-                [0, 0, user32.GetSystemMetrics(0), user32.GetSystemMetrics(1)]]
+def monitorSizes(verbose: int = 0) -> List[List[int]]:
+    """ Return monitor sizes
+
+    Args:
+        verbose: Verbosity level
+    Returns:
+        List with for each screen a list x, y, width, height
+    """
+    _qd = QtWidgets.QDesktopWidget()
```
QFer, line 1766: This line and the check if _qd is None in the next line is not needed
_qd is not used when win32 and in the else _qd is set again in all cases

src/qtt/pgeometry.py
```diff
@@ -1755,37 +1755,38 @@ def ff(xx, figx=fig): return mpl2clipboard(fig=figx)
...
+        import ctypes
+        user32 = ctypes.windll.user32
+        wa = [
+            [0, 0, user32.GetSystemMetrics(0), user32.GetSystemMetrics(1)]]
+    else:
+        _applocalqt = QtWidgets.QApplication.instance()
+
+        if _applocalqt is None:
+            _applocalqt = QtWidgets.QApplication([])
+            _qd = QtWidgets.QDesktopWidget()
```
QFer, line 1777: this line and 1779 are doing the same. Can be out of the condition if: else:

QFer (changes requested):

QFer (approved):

PR #805, opened 2021-12-13
Add qiskit circuit tools

eendebakpt:
* Add method to integrate a circuit into a circuit with more qubits.

### DIFF
--- a/src/qtt/qiskit/circuits.py
+++ b/src/qtt/qiskit/circuits.py
@@ -1,12 +1,13 @@
 from typing import List, Optional
 
+from qiskit.circuit.exceptions import CircuitError
 from qiskit.circuit.quantumcircuit import QuantumCircuit
 
 
 def integrate_circuit(
     qc: QuantumCircuit,
     target_qubits: List[int],
-    target_control_bits: Optional[List[int]] = None,
+    target_classical_bits: Optional[List[int]] = None,
     number_of_qubits: int = 5,
     add_measurements: bool = True,
 ) -> QuantumCircuit:
@@ -18,7 +19,7 @@ def integrate_circuit(
     Args:
         qc: QuantumCircuit to be integrated
         target_qubits: List of qubits to map the circuit on
-        target_control_bits: If None, then use the registers allocated to the target qubits
+        target_classical_bits: If None, then use the registers allocated to the target qubits
         number_of_qubits: Number of qubits in the target system
         add_measurements: If True, then add a measure statement for all the new qubits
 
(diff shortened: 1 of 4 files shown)


src/qtt/qiskit/circuits.py
```diff
@@ -0,0 +1,45 @@
...
+) -> QuantumCircuit:
+    """ Integrate circuit at specified qubits in a larger qubit system
+
+    This can be used for example to integrate a single-qubit experiment in
+    a 5-qubit circuit to be executed on a 5-qubit device.
+
+    Args:
+        qc: QuantumCircuit to be integrated
+        target_qubits: List of qubits to map the circuit on
+        target_control_bits: If None, then use the registers allocated to the target qubits
```
QFer, line 21: These are not control bits but classical bits.

src/qtt/qiskit/circuits.py
```diff
@@ -0,0 +1,45 @@
...
+    output_qc = QuantumCircuit(number_of_qubits, number_of_qubits)
+
+    qbits = [output_qc.qregs[0][i] for i in target_qubits]
+    if target_control_bits is None:
+        cbits = [output_qc.cregs[0][i] for i in target_qubits]
+    else:
+        cbits = target_control_bits
+    output_qc = output_qc.compose(qc, qubits=qbits, clbits=cbits)
+
+    if add_measurements:
```
QFer, line 41: This is a strange option/addition for integration of a circuit, especially when combined more often. Why not keep adding the measurements separate.

src/qtt/qiskit/circuits.py
```diff
@@ -0,0 +1,45 @@
...
+def integrate_circuit(
+    qc: QuantumCircuit,
+    target_qubits: List[int],
+    target_control_bits: Optional[List[int]] = None,
+    number_of_qubits: int = 5,
+    add_measurements: bool = True,
+) -> QuantumCircuit:
+    """ Integrate circuit at specified qubits in a larger qubit system
+
+    This can be used for example to integrate a single-qubit experiment in
```
QFer, line 15: I doubt if this is always logical functionality. When the original circuit has 2-qubit gates and only one of them appears in target_qubits what happens then?

src/qtt/qiskit/circuits.py
```diff
@@ -0,0 +1,45 @@
...
+        raise Exception(
+            f'number of qubits {qc.num_qubits} in the specified circuit cannot exceed specfied number of qubits {number_of_qubits} in the target system')
+
+    output_qc = QuantumCircuit(number_of_qubits, number_of_qubits)
+
+    qbits = [output_qc.qregs[0][i] for i in target_qubits]
+    if target_control_bits is None:
+        cbits = [output_qc.cregs[0][i] for i in target_qubits]
+    else:
+        cbits = target_control_bits
```
QFer, line 38: shouldn't this be cbits = [output_qc.qregs[0][i] for i in target_control_bits]

src/qtt/qiskit/circuits.py
```diff
@@ -0,0 +1,45 @@
+from typing import List, Optional
+
+from qiskit.circuit.quantumcircuit import QuantumCircuit
+
+
+def integrate_circuit(
+    qc: QuantumCircuit,
```
QFer, line 7: What happens when qc exists of more than 1 groups of qregs?

QFer (changes requested):

src/qtt/qiskit/circuits.py
eendebakpt: That is not how the `target_qubits` argument works. The 2-qubit gate in a 3-qubit circuit has qubit indices 1, 2 in the original circuit and the target qubits are 1,3, 4 then the 2-qubit gate is applied on the resulting circuit on qubits 3, 4.

src/qtt/qiskit/circuits.py
eendebakpt: If the original circuit has measurements, then in the constructed circuit there are only measurements on the `target_qubits`. A common operation is then to add measurements to the remaining qubits.

src/qtt/qiskit/circuits.py
eendebakpt: That is no problem. I added a test case to be sure

QFer (approved):

PR #811, opened 2022-01-03
Add method to retrieve current time in measure_time

eendebakpt:


src/qtt/utilities/tools.py
```diff
@@ -69,15 +69,26 @@ class measure_time():
...
     def __init__(self, message: Optional[str] = 'dt: '):
         self.message = message
+        self.dt = -1
 
     def __enter__(self):
         self.start_time = time.perf_counter()
         return self
 
+    @property
+    def current_time(self) -> float:
```
QFer, line 79: When 'delta_time' is the method to get the total time spend in the context, this method/property is better named 'current_delta_time'

QFer (changes requested):

QFer (approved):

PR #812, opened 2022-01-03
Allow hostname to be set in rda_t

peendebak:


src/qtt/gui/live_plotting.py
```diff
@@ -62,48 +68,59 @@ def __init__(self):
...
-            value (int)
+            Value retrieved
         """
         v = self.get(key, default_value)
         if v is None:
             return v
         return int(float(v))
 
-    def get(self, key, default_value=None):
+    def get(self, key: str, default_value: Optional[Any] = None) -> Any:
```
QFer, line 101: The returning type should be Optional[Any]. When value is None and default_value is None (the default value of default value) it returns None.

QFer (changes requested):

src/qtt/gui/live_plotting.py
peendebak: Is this only to make it symmetric with the other return values? Because `Any` includes a `None` right? 

I updated the annotation as suggested

QFer: Optional[T] is just an alias for Union[T, None].
Union[Any, None] isn't the same as Any (https://github.com/python/mypy/issues/3138)

QFer (approved):

PR #813, opened 2022-01-07
Add python 3.10 to CI

eendebakpt:
* Use python 3.10 in CI
* Fix various issues in tests


src/qtt/pgeometry.py
```diff
@@ -37,7 +37,7 @@
 import warnings
 from functools import wraps
 from math import cos, sin
-from typing import Any, List, Optional, Tuple, Union
+from typing import Any, List, Optional, Sequence, Tuple, Union
```
QFer, line 40: Why importing Sequence without using it?

QFer (approved):

PR #817, opened 2022-01-19
Fix scanjob_t parsing for negative scan direction

peendebak:
Minimal example that is fixed with this PR:
```
from qtt.measurements.scans import scanjob_t
import qtt.simulation.virtual_dot_array

station = qtt.simulation.virtual_dot_array.initialize()

num=566
scanjob=scanjob_t({'minstrument': [1], 'minstrumenthandle': ('m4i', [1]), 'wait_time_startscan': 0.04, 'stepdata': {'wait_time': 0.2}, 'Naverage': 2000, 'sweepdata': {'period': 0.0005, 'param': 'P1', 'range': -5, 'step': -0.01098901098901099, 'start': 714.84130859375, 'end': 709.84130859375}, 'scantype': 'scan1Dfast'})
_,s=scanjob._convert_scanjob_vec(station=station, sweeplength=num)
print(s)

assert len(s)==num
```` 
* Fix scanjob_t parsing
* Fix re_arm compensation
* Fix ZI import

### DIFF
--- a/src/qtt/measurements/scans.py
+++ b/src/qtt/measurements/scans.py
@@ -483,10 +483,9 @@ def scan1Dfast(station, scanjob, location=None, liveplotwindow=None, delete=True
 
     is_m4i = _is_m4i(minstrhandle)
     if is_m4i:
-        device_parameters['trigger_re_arm_compensation'] = True
         dt = period*(1-sawtooth_width)/2
         zero_padding = (40+32)/get_sampling_frequency(minstrhandle) - dt
-        print(f'period {period} zero_padding {zero_padding}')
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/qtt/measurements/scans.py [hunk 1]
```diff
@@ -851,7 +850,7 @@ def _calculate_sweepvalues(param, sweep_data, inclusive_end=False) -> SweepFixed
                 raise ValueError('step value may not be 0')
 
             if inclusive_end:
-                epsilon = 1e-8
+                epsilon = np.sign(sweep_data['step'])*1e-8
```
QFer, line 853: Nice catch. It would be great when you add the minimal example given in the description of the pr as an succeeding test in unittests/measurements/test_scans.py

QFer (changes requested):

PR #818, opened 2022-01-20
Add parameter results to plots of RTS and two-level threshold

eendebakpt:
* Add parameters to plots
* Add type annotations

### DIFF
--- a/src/qtt/algorithms/random_telegraph_signal.py
+++ b/src/qtt/algorithms/random_telegraph_signal.py
@@ -151,7 +151,7 @@ def _create_integer_histogram(durations):
 
 
 def tunnelrates_RTS(data: Union[np.ndarray, qcodes.data.data_set.DataSet], samplerate: Optional[float] = None, min_sep: float = 2.0, max_sep: float = 7.0, min_duration: int = 5,
-                    num_bins: Optional[int] = None, fig: Optional[int] = None, ppt=None, verbose: int = 0) -> Tuple[float, float, dict]:
+                    num_bins: Optional[int] = None, fig: Optional[int] = None, ppt=None, verbose: int = 0) -> Tuple[Optional[float], Optional[float], dict]:
     """
     This function takes an RTS dataset, fits a double gaussian, finds the split between the two levels,
     determines the durations in these two levels, fits a decaying exponential on two arrays of durations,
--- a/src/qtt/algorithms/random_telegraph_signal.py
+++ b/src/qtt/algorithms/random_telegraph_signal.py
@@ -5,8 +5,8 @@ Created on Wed Feb 28 10:20:46 2018
 @author: riggelenfv
 """
 
+import operator
 import warnings
-# %%
 from typing import Optional, Tuple, Union
 
 import matplotlib.pyplot as plt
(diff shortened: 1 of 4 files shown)


src/qtt/algorithms/random_telegraph_signal.py
```diff
@@ -73,21 +73,26 @@ def _plot_rts_histogram(data, num_bins, double_gaussian_fit, split, figure_title
     _, bins, _ = plt.hist(data, bins=num_bins)
     bincentres = np.array([(bins[i] + bins[i + 1]) / 2 for i in range(0, len(bins) - 1)])
 
-    plt.plot(bincentres, double_gaussian(bincentres, double_gaussian_fit), 'r', label='Fitted double gaussian')
+    left_gaussian = [double_gaussian_fit[i] for i in [4, 2, 0]]
```
QFer, line 76: The wrong data format is chosen for this structure. These indexes are not clear for the reader. It const me some time to figure out what is happening. Instead of rebuilding we earlier did something like:

    get_left_mean_std_amplitude = operator.itemgetter(4, 2, 0)
    get_right_mean_std_amplitude = operator.itemgetter(5, 3, 1)
    left_gaussian = list(get_left_mean_std_amplitude(double_gaussian_fit))
    right_gaussian = list(get_right_mean_std_amplitude(double_gaussian_fit))

src/qtt/algorithms/random_telegraph_signal.py
```diff
@@ -142,8 +150,8 @@ def _create_integer_histogram(durations):
     return counts, bin_edges, bin_size
 
 
-def tunnelrates_RTS(data, samplerate=None, min_sep=2.0, max_sep=7.0, min_duration=5,
-                    num_bins=None, plungers=None, fig=None, ppt=None, verbose=0):
+def tunnelrates_RTS(data: Union[np.ndarray, qcodes.data.data_set.DataSet], samplerate: Optional[float] = None, min_sep: float = 2.0, max_sep: float = 7.0, min_duration: int = 5,
```
QFer, line 153: Additions are fine, but lines too long

src/tests/unittests/algorithms/test_random_telegraph_signal.py
```diff
@@ -76,3 +78,10 @@ def test_create_integer_histogram(self):
 
         with self.assertRaises(Exception):
             _ = _create_integer_histogram([])
+
+    def test_two_level_threshold(data):
+
+        data = generate_RTS_signal(100000, std_gaussian_noise=0.24, rate_up=10e3, rate_down=20e3)
+        result = two_level_threshold(data)
+        plot_two_level_threshold(result, fig=1)
```
QFer, line 86: Because the figure is printed this one should be added to test_random_telegraph_signal.py in subdirectory integration.

QFer (changes requested):

src/qtt/algorithms/random_telegraph_signal.py
peendebak: That is a reasonable solution yes. Without backwards compatibility we would consider different formats

QFer (approved):

PR #819, opened 2022-01-20
Cleanup of awg_to_plunger

peendebak:
* Cleanup awg_to_plunger code
* Type annotations
* Improve plot titles
* Fix issue with calculation of projective offset rho

### DIFF
--- a/src/qtt/algorithms/awg_to_plunger.py
+++ b/src/qtt/algorithms/awg_to_plunger.py
@@ -155,12 +155,12 @@ def analyse_awg_to_plunger(result: dict, method: str = 'hough', fig: Optional[in
             plt.axis('image')
 
             if angle_pixel is not None:
-                for offset in [-40, -20, 0, 20, 40]:
+                for offset_in_pixels in [-40, -20, 0, 20, 40]:
                     label = None
-                    if offset == 0:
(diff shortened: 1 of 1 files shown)


src/qtt/algorithms/awg_to_plunger.py
```diff
@@ -239,19 +221,19 @@ def plot_awg_to_plunger(result, fig=10):
...
-    plt.figure(fig)
-    plt.clf()
-    MatPlot(ds.default_parameter_array(), num=fig)
+    plot_dataset(ds, fig=fig)
     if angle is not None:
-        rho = -(xscan[0] * np.cos(angle) - np.sin(angle) * xscan[1])
+        rho = -(centre[0] * np.cos(angle) - np.sin(angle) * centre[1])
 
-        for offset in [-40, -20, 0, 20, 40]:
+        for offset in dy*np.linspace(-1, 1, 7):
```
QFer, line 232: Shouldn't this change also be done in line 158 of analyse_awg_to_plunger?

peendebak: No, the image scale is different there. I did rename `offset` to `offset_in_pixels` to make this more clear

QFer (approved):

PR #821, opened 2022-02-07
Update default method for lmfit in gauss_ramsey fitting

peendebak:
- Update default fitting method for lmfit in `fit_gauss_ramsey`
- Add regression tests


src/qtt/algorithms/functions.py
```diff
@@ -359,7 +359,7 @@ def gauss_ramsey_model(x, amplitude, decay_time, frequency, phase, offset):
     lmfit_model.set_param_hint('amplitude', min=0)
     lmfit_model.set_param_hint('decay_time', min=0)
     lmfit_result = lmfit_model.fit(y_data, x=x_data, **dict(zip(lmfit_model.param_names, initial_parameters)),
-                                   verbose=verbose >= 2, weights=weights)
+                                   verbose=verbose >= 2, weights=weights, method='least_squares')
```
QFer, line 362: We did this change earlier in fitting.py
Now in qtt only one instance is left with default method 'leastsq'. In fit_sine in algorithm/fitting.py
Please check if we need to change the method there too.

Further the change looks good and can be approved after this check

peendebak: Done

QFer (approved):

PR #825, opened 2022-02-10
Remove double plotting of initial estimate in AllXY

eendebakpt:
* The initial estimate was plotted twice, this is removed
* Add dataset location to plot title


QFer (approved):

QFer:
@eendebakpt We can add this one to the release if you rebase.

PR #831, opened 2022-03-07
Refactor markov chain code

eendebakpt:
* Add type annotations
* Fix spelling mistakes
* Convert `ChoiceGenerator` to dataclass

### DIFF
--- a/src/qtt/algorithms/markov_chain.py
+++ b/src/qtt/algorithms/markov_chain.py
@@ -198,7 +198,7 @@ class ContinuousTimeMarkovModel:
             initial_state = self.stationary_distribution().flatten().tolist()
             initial_state = random.choices(range(number_of_states), weights=initial_state, k=1)[0]  # type: ignore
         elif isinstance(initial_state, (list, np.ndarray, tuple)):
-            initial_state = np.asarray(initial_state).tolist()
+            initial_state = np.asarray(initial_state).flatten().tolist()
             initial_state = random.choices(range(number_of_states), weights=initial_state, k=1)[0]  # type: ignore
 
         P = self.transition_matrix(delta_time)
@@ -230,7 +230,7 @@ class ContinuousTimeMarkovModel:
             Array with generated sequences
         """
         if initial_state is None:
-            initial_state = self.stationary_distribution().tolist()
+            initial_state = self.stationary_distribution()
         sequences = np.zeros((number_of_sequences, length), dtype=int)
         for n in range(number_of_sequences):
             sequences[n] = self.generate_sequence(length, delta_time, initial_state)
--- a/src/qtt/algorithms/markov_chain.py
+++ b/src/qtt/algorithms/markov_chain.py
@@ -38,7 +38,8 @@ class ChoiceGenerator:
     def __post_init__(self):
         if not self.number_of_states == len(self.cum_weights):
             raise Exception(
-                f'specification of cumulative weights {len(self.cum_weights)} does not match number of states {self.number_of_states}')
+                f'specification of cumulative weights (len {len(self.cum_weights)})'
+                + ' does not match number of states {self.number_of_states}')
         self._idx = 0
         self._block = self._generate_block()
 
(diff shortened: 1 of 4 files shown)


src/qtt/algorithms/markov_chain.py
```diff
@@ -1,74 +1,88 @@
...
-            raise Exception('specification of cummulative weights does not match number of states')
+    number_of_states: int
+    cum_weights: np.ndarray
+    block_size: int = 5000
 
-        self.number_of_states = number_of_states
+    def __post_init__(self):
+        if not self.number_of_states == len(self.cum_weights):
+            raise Exception(
+                f'specification of cumulative weights {len(self.cum_weights)} does not match number of states {self.number_of_states}')
```
QFer, line 41: line too long

src/qtt/algorithms/markov_chain.py
```diff
@@ -1,74 +1,88 @@
...
 
         Args:
-            states (str[]): list with names for the states
-            holding_parameters(float[]): List with the holding parameters. The holding parameters determine the average
+            states: list with names for the states
+            holding_parameters: List with the holding parameters. The holding parameters determine the average
                 time before the system will make a jump to a new state
-            jump_chain (array): The jump chain or transition matrix. This matrix gives the probability for the system
+            jump_chain: The jump chain or transition matrix. This matrix gives the probability for the system
                         to jump from a state to one of the other states. The sum of the probabilities in each column must
```
QFer, line 86: line too long - 1 char

src/qtt/algorithms/markov_chain.py
```diff
@@ -1,74 +1,88 @@
...
-        """ Generate a choice
+    def generate_choice(self) -> int:
+        """ Generate a single choice
 
         Returns:
-            int: integer in the range 0 to the number of states
+            Integer in the range 0 to the number of states
         """
         self._idx = self._idx + 1
-        if self._idx == self._block_size:
```
QFer, line 56: generate_choice doesn't return item 0 for the first generated block. In line 42, self._idx is initialized to 0 and here it is increased before returning the element. while after generating the second block the first item return has self._idx = 0.
I don't know exactly how this is used if this is an issue

src/qtt/algorithms/markov_chain.py
```diff
@@ -1,74 +1,88 @@
...
+    number_of_states: int
+    cum_weights: np.ndarray
+    block_size: int = 5000
 
-        self.number_of_states = number_of_states
+    def __post_init__(self):
+        if not self.number_of_states == len(self.cum_weights):
+            raise Exception(
+                f'specification of cumulative weights {len(self.cum_weights)} does not match number of states {self.number_of_states}')
         self._idx = 0
```
QFer, line 42: __post_init__ shouldn't add instance attributes (self._block and self._idx). An __init__ should do that or why not add them as (initialized) instance attributes just as block_size.

src/qtt/algorithms/markov_chain.py
```diff
@@ -163,54 +177,57 @@ def stationary_distribution_discrete(jump_chain):
...
-            array : generated sequence
+            Array with generated sequence
 
         """
         number_of_states = self.number_of_states()
         if initial_state is None:
-            initial_state = self.stationary_distribution()
-            initial_state = random.choices(range(number_of_states), weights=initial_state, k=1)[0]
+            initial_state = self.stationary_distribution().flatten().tolist()
+            initial_state = random.choices(range(number_of_states), weights=initial_state, k=1)[0]  # type: ignore
```
QFer, line 199: is mypy happy with this reuse of initial_state?

QFer (changes requested):

src/qtt/algorithms/markov_chain.py
eendebakpt: This is no issue, it just means that we generate one random number that is not used.  We could avoid it, but it would  make the rest of the flow of the nice a bit more involved.

src/qtt/algorithms/markov_chain.py
eendebakpt: mypy is not happy because `random.choices` returns a list without type annotation. the input argument to `random.choices` is `range(number_of_states)` we we know it is an int

src/qtt/algorithms/markov_chain.py
eendebakpt: The `_block` and `_idx` are private internal variables. I do not want them in the representation of the object, and also not in the data when the dataclass is converted to json or a dictionary, which is why they are not defined as the `block_size`.
`__init__` is something one should not define for a dataclass

QFer (approved):

PR #835, opened 2022-03-11
[Performance] Simplify polyarea

peendebak:
* The `polyarea` method contained a code path for the numpy array and for the list type. This PR simplifies the method to only have a single implementation. The `zip` based approach is faster. For the example contained in the docstring the `zip` based method averaged to 2 us, while the numpy method averages to 16 us.

* Improve performance of directionMean and circular_mean

* Add some type annotations and improve docstrings

### DIFF
--- a/src/qtt/pgeometry.py
+++ b/src/qtt/pgeometry.py
@@ -36,6 +36,7 @@ import time
 import warnings
 from functools import wraps
 from typing import Any, List, Optional, Tuple, Union
+from math import cos, sin
 
 import numpy
 import numpy as np
@@ -168,9 +169,8 @@ def memory() -> float:
     """ Return the memory usage in MB
 
     Returns:
-            Memory usage in MB
+          Memory usage in MB
     """
-    import os
 
     import psutil
     process = psutil.Process(os.getpid())
@@ -271,7 +271,7 @@ def static_var(varname: str, value: Any):
 
 
 @static_var("time", {'default': 0})
-def tprint(string: str, dt: float = 1, output: bool = False, tag: str = 'default'):
+def tprint(string: str, dt: float = 1, output: bool = False, tag: str = 'default') -> Optional[bool]:
     """ Print progress of a loop every dt seconds
 
     Args:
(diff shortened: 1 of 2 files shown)


src/qtt/pgeometry.py
```diff
@@ -665,21 +665,25 @@ def callbackF(Xi):
...
-    r = scipy.optimize.minimize(cost_function, a0, callback=None, options=dict({'disp': False}))
+    r = scipy.optimize.minimize(cost_function, angle_initial_guess, callback=None, options=dict({'disp': False}))
     angle = r.x[0]
     return angle
 
 
 def circular_mean(weights, angles):
     """ Calculate circular mean of a set of 2D vectors """
     x = y = 0.
+    cos = math.cos
```
QFer, line 681: why not from math import cos, sin, radians. There are more instances of math.cos etc. in this modules not touched.

src/qtt/pgeometry.py
```diff
@@ -270,40 +270,36 @@ def decorate(func):
 
 
 @static_var("time", {'default': 0})
-def tprint(string, dt=1, output=False, tag='default'):
+def tprint(string: str, dt: float = 1, output: bool = False, tag: str = 'default') -> Union[bool, None]:
```
QFer, line 273: return type is -> Optional[bool]

src/qtt/pgeometry.py
```diff
@@ -270,40 +270,36 @@ def decorate(func):
...
-    if (time.time() - tprint.time.get(tag, 0)) > dt:
+    if (time.perf_counter() - tprint.time.get(tag, 0)) > dt:
         print(string)
-        tprint.time[tag] = time.time()
+        tprint.time[tag] = time.perf_counter()
         if output:
             return True
         else:
-            return
+            return None
```
QFer, line 291: Optionally this can be changed to:

```
    if (time.perf_counter() - tprint.time.get(tag, 0)) > dt:
        print(string)
        tprint.time[tag] = time.perf_counter()
        if output:
            return True
    elif output:
        return False
    return None
```

src/qtt/pgeometry.py
```diff
@@ -665,21 +665,25 @@ def callbackF(Xi):
         Nfeval += 1
 
     m = vec.mean(axis=0)
-    a0 = np.arctan2(m[0], m[1])
+    angle_initial_guess = np.arctan2(m[0], m[1])
 
-    def cost_function(a): return dist(a, vec)
+    def cost_function(a):
+        return dist(a, vector_angles)
```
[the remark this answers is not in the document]
peendebak: Doesn't make sense indeed. The callbackF should have been removed. Not sure why it was still there.

QFer (approved):

PR #838, opened 2022-03-18
improve performance by combining calls to percentile

eendebakpt:
For combined calls the sorting inside `percentile` only needs to be done once.

@QFer


src/qtt/algorithms/coulomb.py
```diff
@@ -79,13 +76,14 @@ def fitCoulombPeaks(x_data, y_data, lowvalue=None, verbose=1, fig=None, sampling
...
-    minval = np.percentile(y_data, 5) + 0.1 * (np.percentile(y_data, 95) - np.percentile(y_data, 5))
+    p1, p5, p95, p99 = np.percentile(y_data, [1, 5, 95, 99])
+    minval = p5 + 0.1 * (p95 - p5)
     local_maxima, _ = nonmaxsuppts(y_data, d=int(12 / sampling_rate), minval=minval)
     fit_data = fitPeaks(x_data, y_data, local_maxima, verbose=verbose >= 2, fig=fig)
 
     if lowvalue is None:
-        lowvalue = np.percentile(y_data, 1)
-    highvalue = np.percentile(y_data, 99)
+        lowvalue = p1
```
QFer, line 85: lowvalue parameter should be added to the docstring, explaining this default value when None.

src/qtt/algorithms/coulomb.py
```diff
@@ -659,8 +656,8 @@ def findBestSlope(x, y, minimal_derivative=None, fig=None, verbose=1):
         slopes (...)
         results (object): additional data
     """
-    lowvalue = np.percentile(y, 1)
-    highvalue = np.percentile(y, 99)
+    lowvalue, highvalue = np.percentile(y, [5, 99])
```
QFer, line 659: Are you aware you changed the lowvalue to 5th percentile instead of the first?

QFer (changes requested):

src/qtt/algorithms/coulomb.py
peendebak: Good spot

QFer (approved):

PR #841, opened 2022-04-05
Optimize choice generation

eendebakpt:


src/qtt/algorithms/markov_chain.py
```diff
@@ -59,10 +62,10 @@ def generate_choice(self) -> int:
             Integer in the range 0 to the number of states
         """
         self._idx = self._idx + 1
-        if self._idx == self._block.size:
+        if self._idx == self.block_size:
             self._idx = 0
-            self._block = self._generate_block()
-        return int(self._block[self._idx])
+            self._block = self._generate_block().tolist()
```
QFer, line 67: To see this working (twice) the unit test `test_ChoiceGenerator` in test_markov_chain.py, line 17 should be range(10000) (bigger than default block_size 5000)

src/qtt/algorithms/markov_chain.py
```diff
@@ -179,7 +182,8 @@ def stationary_distribution_discrete(jump_chain) -> np.ndarray:
         return pi
 
     def generate_sequence(self, length: int, delta_time: float,
-                          initial_state: Union[None, int, np.ndarray] = None) -> np.ndarray:
+                          initial_state: Union[None, int, np.ndarray] = None,
+                          generators: Optional[List[ChoiceGenerator]] = None) -> np.ndarray:
```
QFer, line 186: This parameter generators is added but to feed this generate sequence with an alternative generator than the default is difficult, it is tightly coupled to this class.
It is not wrong but I think it will not be used. Do you see a use case for it?

QFer (changes requested):

src/qtt/algorithms/markov_chain.py
```diff
@@ -202,20 +207,24 @@ def generate_sequence(self, length: int, delta_time: float,
...
 
-        P = self.transition_matrix(delta_time)
+        if generators is None:
+            generators = self._create_generators(delta_time)
 
         sequence = np.zeros(length, dtype=int)
-        sequence[0] = initial_state
+        sequence[0] = value = initial_state
+        for i in range(1, sequence.size):
+            sequence[i] = value = generators[value].generate_choice()
```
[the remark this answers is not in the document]
peendebak: The original code was indeed clearer, but this is faster (and in the hot path of the calculations). I added a comment on this

src/qtt/algorithms/markov_chain.py
peendebak: I agree that it won't be used. It was added to eliminate the overhead of constructing generators on the fly. I considered making the generators a private attribute of the structure, but then we would have to check whether those generators match with the `delta_time` parameter.
If a future PR we could clean this up a bit and make the attributes like `jump_chain` read only.

QFer (approved):

PR #845, opened 2022-04-13
Remove legacy virtual_awg instrument

eendebakpt:


QFer (changes requested):
This file is still used in:
docs\notebooks\unfinished\tektronix_virtual_awg_example.py 

src\qtt\exampledata\
also a number of example data snapshots (json) refer to this file (instrument)

"instrument": "qtt.instrument_drivers.virtual_awg.virtual_awg"

eendebakpt:
> 

1. and 3. done. 2. is not needed, there are many instruments in the snapshot, but they are not used

QFer (approved):

PR #847, opened 2022-04-14
Add method to determine tunnel rate ratio 

eendebakpt:
* Add method to determine tunnel rate ratio based only on mean and std dev of signal
* Fix minor inconsistency in calculation of thresholded data


src/qtt/algorithms/random_telegraph_signal.py
```diff
@@ -19,7 +19,33 @@
...
-# %% calculate durations of states
+
+def rts2tunnel_ratio(binary_signal: np.ndarray) -> float:
+    """ Calculate ratio between tunnelrate down and up
+
+    From the mean and standard deviation of the RTS data we can determine the ratio between
+    the two tunnel rates. See equations on https://en.wikipedia.org/wiki/Telegraph_process
+
+    Args:
+        binary_signal: RTS signal with two levels
```
QFer, line 30: also the levels must be 0 and 1 (or False and True)

src/qtt/algorithms/random_telegraph_signal.py
```diff
@@ -19,7 +19,33 @@
...
+
+    Returns:
+        Ratio of tunnelrate up to down and down to up
```
QFer, line 33:
```suggestion
        Ratio of tunnelrate up to down (l2) and down to up (l1)
```

src/qtt/algorithms/random_telegraph_signal.py
```diff
@@ -19,7 +19,33 @@
...
+    c2 = np.max(binary_signal)
+    if c1 == c2:
+        raise Exception(f'binary signal contains only a single value {c1}')
+
+    if c1 != 0 or c2 != 1:
+        raise NotImplementedError('signal must only contain 0 and 1')
+    m = np.mean(binary_signal)
+    var = np.var(binary_signal)
+
+    ratio_l2_over_l1 = var/m**2
```
QFer, line 46: When we only have a l2 transition (and no l1), we get outcome 1 is that oke?
binary_signal = [True, True, False, False]

QFer (changes requested):

QFer (approved):

PR #848, opened 2022-04-19
RTS improvements

eendebakpt:
* improve performance of xdata definition
* improve performance of fitting and plotting for large datasets


src/qtt/algorithms/random_telegraph_signal.py
```diff
@@ -243,9 +243,10 @@ def tunnelrates_RTS(data: Union[np.ndarray, qcodes.data.data_set.DataSet], sampl
 
     # plotting a 2d histogram of the RTS
     if fig:
-        xdata = np.array(range(0, len(data))) / samplerate * 1000
-        Z, xedges, yedges = np.histogram2d(xdata, data, bins=[int(
-            np.sqrt(len(xdata)) / 2), int(np.sqrt(len(data)) / 2)])
+        xdata = np.arange(len(data)) / (samplerate / 1000)
+        ny = min(int(np.sqrt(len(data))/2), 800)
+        nx = min(int(np.sqrt(len(data))/2), 1200)
```
QFer, line 248: data should be xdata

src/qtt/algorithms/random_telegraph_signal.py
```diff
@@ -243,9 +243,10 @@ def tunnelrates_RTS(data: Union[np.ndarray, qcodes.data.data_set.DataSet], sampl
 
     # plotting a 2d histogram of the RTS
     if fig:
```
QFer, line 245:
```suggestion
    if fig is not None:
```

src/qtt/algorithms/random_telegraph_signal.py
```diff
@@ -260,7 +261,7 @@ def tunnelrates_RTS(data: Union[np.ndarray, qcodes.data.data_set.DataSet], sampl
 
     # binning the data and determining the bincentres
     if num_bins is None:
-        num_bins = int(np.sqrt(len(data)))
+        num_bins = min(int(np.sqrt(len(data))), 1200)
```
QFer, line 264: earlier, creating the figure, data array was binned to a max of 800, now 1200?

QFer (changes requested):

src/qtt/algorithms/random_telegraph_signal.py
peendebak: xdata and data have the same size, so it is ok. Will change it anyway

src/qtt/algorithms/random_telegraph_signal.py
peendebak: The binning is different. In this fist case it is for plotting a 2D histogram, for the second case for fitting data. I made the 1200 an explicit variable to clarify. The exact amounts are not really important for these high numbers though.

QFer (approved):

PR #851, opened 2022-05-04
Add raised cosine function

eendebakpt:
Add raised cosine function

### DIFF
--- a/src/qtt/algorithms/functions.py
+++ b/src/qtt/algorithms/functions.py
@@ -471,3 +471,60 @@ def logistic(x, x0=0, alpha=1):
     """
     f = 1 / (1 + np.exp(-2 * alpha * (x - x0)))
     return f
+
+
+def raised_cosine_frequency_domain(f: np.ndarray, beta: float, T: float) -> np.ndarray:
+    """ Raised cosine frequency domain function
+
+    See https://en.wikipedia.org/wiki/Raised-cosine_filter
+
+    Args:
+        t: Frequency
+        beta: Roll-off factor
+        T: Symbol period
+
+    Returns:
+        Calculated values of the raised cosine
+    """
+    two_T = 2*T
+    if beta == 0:
+        return np.abs(f) < 1/(two_T)
+    ww = np.abs(f)-(1-beta)/(two_T)
+    value = .5*(1+np.cos((np.pi*T/beta)*ww))
+    idx = np.abs(f) <= (1-beta)/two_T
+    value[idx] = 1
+    idx = np.abs(f) >= (1+beta)/two_T
+    value[idx] = 0
+    return value
+
+
+def raised_cosine(t: np.ndarray, beta: float, T: float) -> np.ndarray:
+    """ Raised cosine impulse response function
+
+    See https://en.wikipedia.org/wiki/Raised-cosine_filter
+
+    Args:
+        t: Independent variable
+        beta: Roll-off factor
+        T: Symbol period
+
+    Returns:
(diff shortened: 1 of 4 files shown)


src/qtt/algorithms/functions.py
```diff
@@ -471,3 +471,60 @@ def logistic(x, x0=0, alpha=1):
...
+    ww = np.abs(f)-(1-beta)/(two_T)
+    value = .5*(1+np.cos((np.pi*T/beta)*ww))
+    idx = np.abs(f) <= (1-beta)/two_T
+    value[idx] = 1
+    idx = np.abs(f) >= (1+beta)/two_T
+    value[idx] = 0
+    return value
+
+
+def raised_cosine(t: np.ndarray, beta: float, T: float) -> np.ndarray:
```
QFer, line 501: Arguments t, beta and T are not descriptive to me. Especially t and T as parameters of one method. I understand they come from the formulas on wikipedia but we should create descriptive names. Also python doesn't prescribe capitals in argument names.

src/qtt/algorithms/functions.py
```diff
@@ -471,3 +471,60 @@ def logistic(x, x0=0, alpha=1):
     """
     f = 1 / (1 + np.exp(-2 * alpha * (x - x0)))
     return f
+
+
+def raised_cosine_frequency_domain(f: np.ndarray, beta: float, T: float) -> np.ndarray:
```
QFer, line 476: same as the other method, non-descriptive parameter names

src/qtt/algorithms/functions.py
```diff
@@ -471,3 +471,60 @@ def logistic(x, x0=0, alpha=1):
...
+    Returns:
+        Calculated values of the raised cosine
+    """
+    two_T = 2*T
+    if beta == 0:
+        return np.abs(f) < 1/(two_T)
+    ww = np.abs(f)-(1-beta)/(two_T)
+    value = .5*(1+np.cos((np.pi*T/beta)*ww))
+    idx = np.abs(f) <= (1-beta)/two_T
+    value[idx] = 1
```
QFer, line 495: same as a later comment, idx is boolean value. I guess you use implicit casting but even then for me it is strange to read.

idx = 1 if np.abs(f) <= (1-beta)/two_T else 0

something like this?

QFer (changes requested):

src/qtt/algorithms/functions.py
```diff
@@ -471,3 +471,60 @@ def logistic(x, x0=0, alpha=1):
...
+    Returns:
+        Calculated values of the raised cosine
+    """
+    t = np.asarray(t)
+    t_div_T = t / T
+    if beta == 0:
+        return (1 / T) * np.sinc(t_div_T)
+
+    idx = np.abs(t) == T / (2 * beta)
+    t_div_T[idx] = 0
```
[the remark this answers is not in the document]
eendebakpt: `idx`  is a boolean array. It is normal to index numpy arrays using such an object, see https://numpy.org/devdocs/user/basics.indexing.html#boolean-array-indexing

src/qtt/algorithms/functions.py
eendebakpt: Same as above. Note that scalar input values are not allowed here, so `raised_cosine(1.1, .2, .4)` will fail. (for this reason the argument is annotated as `np.ndarray`, and not ` npt.ArrayLike`)

QFer (approved):

PR #870, opened 2023-01-31
Fix/op nnn/python 3 10 update

QFer:
* Added updates to dev
* Fixed mypy and some tests
* Changelog adjusted


setup.py
```diff
@@ -35,17 +35,17 @@ def get_package_data(root_dir):
 tests_require = ['coverage', 'jupyter', 'mypy', 'types-redis', 'types-python-dateutil', 'pytest']
 
 install_requires = [
-    'apscheduler', 'attrs', 'dulwich', 'IPython>=0.1', 'jupyter', 'lmfit', 'matplotlib>=3.0',
-    'numdifftools', 'numpy>=1.15', 'opencv-python', 'pandas', 'PyQt5', 'pyqtgraph', 'pyvisa', 'pyzmqrpc', 'qcodes>=0.23.0',
-    'qcodes-contrib-drivers', 'qilib', 'qiskit-terra>=0.19', 'qtpy', 'qupulse', 'redis', 'scipy', 'scikit-image',
-    'shapely', 'sympy', 'imageio'
+    'apscheduler', 'attrs', 'dulwich', 'IPython', 'jupyter', 'lmfit', 'matplotlib',
+    'numdifftools', 'numpy>=1.23', 'opencv-python', 'pandas', 'PyQt5', 'pyqtgraph<0.13.0', 'pyvisa', 'pyzmqrpc', 'qcodes>=0.30.0',
```
eendebakpt, line 39:
```suggestion
    'numdifftools', 'numpy>=1.23', 'opencv-python', 'pandas', 'PyQt5', 'pyqtgraph', 'pyvisa', 'pyzmqrpc', 'qcodes>=0.30.0',
```
This should not be necessary any longer with recent qcodes. I can also do this in a followup PR.

.readthedocs.yml
```diff
@@ -1,8 +1,8 @@
 build:
   image: latest
-  
+
 python:
-  version: 3.8
+  version: 3.7
```
eendebakpt, line 5: Python 3.7 has basically reached its end of life: https://endoflife.date/python. I think 3.9 or 3.10 would be better here (depending a bit on what RTD offers)

src/qtt/qi.py
```diff
@@ -0,0 +1,49 @@
+"""
```
eendebakpt, line 1: I think we can remove this, I will keep the code locally.

eendebakpt (approved):

.readthedocs.yml
QFer: Agree, we can increase the version to 3.10 or 3.9 in a next pr.

setup.py
QFer: Ok. we should also increase the base python version for qtt to 3.8 soon?

PR #872, opened 2023-03-09
Fix awg_to_plunger

eendebakpt:
The example at https://qtt.readthedocs.io/en/latest/notebooks/analysis/example_awg_to_plunger.html shows incorrect results. This PR fixes the issue and refactors the code a bit.

@QFer


QFer (commented):
Would be great when you add a unit test where the awg_to_plunger failed to produce the right results based on the old code.

QFer (approved):

PR #873, opened 2023-03-17
Fix hom method

eendebakpt:
The `qtt.pgeometry.hom` method contained a bug introduced in refactoring.

### DIFF
--- a/src/tests/unittests/test_pgeometry.py
+++ b/src/tests/unittests/test_pgeometry.py
@@ -4,7 +4,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 
 import qtt.pgeometry as pgeometry
-from qtt.pgeometry import pg_transl2H, point_in_polygon, points_in_polygon, projectiveTransformation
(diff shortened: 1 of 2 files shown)


src/tests/unittests/test_pgeometry.py
```diff
@@ -18,6 +18,13 @@ def test_robust_cost():
 
 class TestGeometryOperations(unittest.TestCase):
 
+    def test_hom(self):
+        pts = np.array([[1,0], [1,1], [2,2]]).T
+        expected = np.array([[1, 1, 2],
+               [0, 1, 2],
+               [1, 1, 1] ])
+        self.assertEqual(hom(pts), expected)
```
QFer: Test fails in the pipeline

QFer (changes requested):

src/tests/unittests/test_pgeometry.py
eendebakpt: @QFer Fixed

QFer (approved):

PR #877, opened 2023-05-01
Fix QTT for QCodes 0.38

bkreynen:
In the release of QCodes 0.38 some things were split into qcodes and qcodes_loop this caused deprecation warnings and errors in QTT. This PR fixes those warnings and errors.


src/qtt/__init__.py
```diff
@@ -133,20 +133,17 @@ def _redisStrSet(value, var='qtt_live_value1'):
 
 abort_measurements = _abort_measurement  # type: ignore
 
-# patch the qcodes abort function
```
bkreynen, line 136: This was removed altogether because QCodes does not have a abort_measurements function to override (anymore?) and neither does qcodes_loops. It was difficult to determine whether QCodes did have it back then since this line was added when QCodes 0.1.0 was still used and it is unclear from the GitHub repository of QCodes which version of the source code was used to build the 0.1.0 release.

eendebakpt (approved):

PR #881, opened 2023-07-06
Fix cv2 mypy errors in pipeline

bkreynen:
A new release of opencv-python caused mypy to fail in the pipeline.

These changes should fix the pipeline and should also work with the old version of opencv-python.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ def get_package_data(root_dir):
 tests_require = ['coverage', 'jupyter', 'mypy', 'types-redis', 'types-python-dateutil', 'pytest']
 
 install_requires = [
-    'apscheduler', 'attrs', 'dulwich', 'IPython', 'jupyter', 'lmfit', 'matplotlib',
+    'apscheduler', 'attrs', 'dulwich', 'IPython', 'jupyter', 'lmfit', 'matplotlib<3.8.0',
     'numdifftools', 'numpy>=1.23', 'opencv-python', 'pandas', 'PyQt5', 'pyqtgraph', 'pyvisa', 'pyzmqrpc', 'qcodes>=0.30.0',
     'qcodes-contrib-drivers', 'qcodes-loop', 'qilib==0.3.12', 'qiskit-terra>=0.22', 'qtpy', 'qupulse', 'redis', 'scipy', 'scikit-image',
     'shapely', 'sympy', 'imageio', 'rich'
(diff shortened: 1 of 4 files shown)


src/qtt/pgeometry.py
```diff
@@ -195,12 +195,12 @@ def list_objects(objectype=None, objectclassname='__123', verbose=1):
 
 def package_versions(verbose: int = 1):
     """ Report package versions installed """
-    print('numpy.__version__ %s' % numpy.__version__)
+    print('numpy.__version__ %s' % numpy)
```
eendebakpt: This does not print the version any longer right? (only the package location). I think numpy.__version__ should be available

bkreynen: You're right, I removed it by mistake. I pushed a new commit that restores it.

eendebakpt (dismissed):

setup.py
```diff
@@ -35,7 +35,7 @@ def get_package_data(root_dir):
 tests_require = ['coverage', 'jupyter', 'mypy', 'types-redis', 'types-python-dateutil', 'pytest']
 
 install_requires = [
-    'apscheduler', 'attrs', 'dulwich', 'IPython', 'jupyter', 'lmfit', 'matplotlib',
+    'apscheduler', 'attrs', 'dulwich', 'IPython', 'jupyter', 'lmfit', 'matplotlib<3.8.0',
```
eendebakpt: @bkreynen Restricting matplotlib does not work well with other package requirements. Why did you pin matplotlib? In the builds I could find https://app.travis-ci.com/github/QuTech-Delft/qtt/builds/265966621 which has some mypy errors. 

Can we fix the mypy errors (or just ignore them, sometimes mypy cannot handle the annotations properly)

bkreynen: I haven't finished working on this yet. I've marked the PR as draft again. I temporarily pinned matplotlib just to check that the new version of matplotlib was the cause of these errors.

bkreynen:
@eendebakpt I've briefly looked into fixing the mypy issues, but this does not seem trivial in this case, so I ignored them. I also looked into the issue with coveralls. It seems the issue with coveralls is that the repository on coveralls was created with a different account than you are using now (peendebak) which no longer is a member of the QuTech-Delft group on Github. The options to fix this are:

1. Delete the repository on coveralls and recreate it with a new account. This should fix the issue but means we loose historical data.
2. Contact coveralls support and ask them to change the owner.

But I am not sure how important it is to you that this check remains working on this repo since from what I've heard no new features are being actively developed QTT. For now the check is no longer required so MRs can be merged without it.

eendebakpt (approved):

PR #882, opened 2023-09-01
Fix bug in subpixelmax

eendebakpt:
Fix bug in subpixelmax where some maxima are flat and others not

@bkreynen


eendebakpt:
Closing as no further updates to qtt will be done

